PR #2, opened 2022-06-09
Adds argparse to enable log to file

jmmb13:
It is not enabled by default but can be for the development phase. I didn't add the debug variable because it is got through the environment variable. Also, I didn't add a version because it required adding the version in a constant or main file, and I didn't know if it was wanted.


balloob:
How do you plan on using this for development? Output is currently printed on the terminal, right where you can read it?

jmmb13:
I'm using the core in my host, and when I start the matter server I append -v to the pyrhon3 matter_server.server command.
Yesterday I had a problem when commissioning the device but because I had a lot of MINMDNS spam from my others smart home devices I couldn't see the logs, I used 2> log_file but I was thinking I could add argparse. I can change de code if it is missing something.

agners:
I think its nice to have the option to generate log files, but by default we should log to the terminal, as this is how it will get captured also by the Matter Server add-on.

So, I'd make log file output dependent of `--log-file` argument. Only write to the log file if that is passed.

The `--log-level` seems like a good idea. Not sure if `--verbose` is needed if we have that, I don't have a strong opinion, if you prefer to keep it as a "quick" option, why not.

jmmb13:
I put --log-evel but I can put --loglevel and remove also the small -log. I defaulted to info for the code have the same behaviour.

agners (changes requested):
You also need to use `log_level` then in some places. Other than that looks good to me. Ideally we should then add log level as a configuration option for the add-on.

agners (changes requested):
This still feels a bit wrong to me: Currently when using `python3 -m matter_server.server --log-file my.log` it won't write logs to the file, only after adding `-v` it will log. This is not how I would expect it to behave.

IMHO, the `-v` can be as a shorthand for `--log-file debug`, but other than that, it should not influence logging.

`--log-file` should be the only flag to influence if a file handler is used or not

`--log-level` should influence the log level.

jmmb13:
You are right, I removed the -v arg.I think right now it has the behavior expected.

matter_server/server/__main__.py
```diff
@@ -9,12 +10,41 @@
...
-logging.basicConfig(level=logging.WARN)
 _LOGGER = logging.getLogger(__package__)
-_LOGGER.setLevel(logging.DEBUG)
+
+
+def get_arguments() -> argparse.Namespace:
+    """Get parsed passed in arguments."""
+
+    parser = argparse.ArgumentParser(
+        description=" Matter Controller Server using WebSockets."
```
agners, line 20:
```suggestion
        description="Matter Controller Server using WebSockets."
```

matter_server/server/__main__.py
```diff
@@ -9,12 +10,41 @@
...
+    return arguments
 
 
 def main() -> int:
+    args = get_arguments()
+
+    if args.log_file:
+        file_handler = logging.FileHandler(args.log_file)
+
+    logging.basicConfig(handlers=[file_handler], level=args.log_level.upper())
```
agners, line 46: This currently calls `basicConfig` with unassigned `file_handler`. We should set handlers to None to let basicConfig create the default `StreamHandler` by default, e.g.

```suggestion
    handlers = None

    if args.log_file:
        handlers = [logging.FileHandler(args.log_file)]

    logging.basicConfig(handlers=handlers, level=args.log_level.upper())
```

agners (changes requested):

matter_server/server/__main__.py
jmmb13: right! Missed that

agners (approved):

PR #17, opened 2022-06-17
Synchronize attribute reads and subscriptions on startup

agners:
Avoid multiple entites requesting reads and subscriptions of the same
node. Nodes seem to be overwhelmed and return a BUSY error code, which
isn't handled gracefully throughout the stack (it seems).

### DIFF
--- a/matter_server/client/model/node.py
+++ b/matter_server/client/model/node.py
@@ -1,5 +1,6 @@
 """Matter node."""
 from __future__ import annotations
+import asyncio
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


matter_server/client/model/node.py
```diff
@@ -16,6 +17,7 @@ class MatterNode:
     """Matter node."""
 
     root_device: MatterDevice[device_types.RootNode]
+    node_lock = asyncio.Lock()
```
balloob, line 20: You need to initialize it in the constructor to make it unique per Node.

balloob (approved):

matter_server/client/model/node.py
balloob: One that that is meh here is that this ties the lock to the agnostic model. I wonder if instead we should store this on the HA side?

balloob (approved):

PR #34, opened 2022-06-19
Set entity description

balloob:
We allowed storing but didn't set the entity description

### DIFF
--- a/custom_components/matter_experimental/entity.py
+++ b/custom_components/matter_experimental/entity.py
@@ -23,7 +23,8 @@ class MatterEntity(entity.Entity):
     def __init__(self, device: MatterDevice, mapping: DeviceMapping) -> None:
         self._device = device
         self._device_mapping = mapping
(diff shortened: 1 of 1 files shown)


custom_components/matter_experimental/entity.py
```diff
@@ -23,6 +23,7 @@ class MatterEntity(entity.Entity):
     def __init__(self, device: MatterDevice, mapping: DeviceMapping) -> None:
         self._device = device
         self._device_mapping = mapping
+        self._entity_description = mapping.entity_description
```
agners, line 26: The attribute seems to be without underline (`self.entity_description`). With that Sensor entities look much better now. It seems that some mappings (e.g. Light) don't have that attribute set, leading to `NoneType` access. So it should be set conditionally:

```suggestion
        if mapping.entity_description:
            self.entity_description = mapping.entity_description
```

agners (approved):

PR #35, opened 2022-06-19
Do not timeout on subscribe/read operations

agners:
Sometimes read/subscribe operations take longer (presumably mainly at
startup while searching for the node and establishing CASE sessions).
The stack has its own timeout handling as well, we should rely on those.

### DIFF
--- a/custom_components/matter_experimental/entity.py
+++ b/custom_components/matter_experimental/entity.py
@@ -8,6 +8,7 @@ from typing import Any, Callable, Coroutine
 import async_timeout
 from homeassistant.core import callback
 from homeassistant.helpers import device_registry, entity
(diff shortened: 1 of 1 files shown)


custom_components/matter_experimental/entity.py
```diff
@@ -59,20 +58,20 @@ async def init_matter_device(self) -> None:
 
         try:
             # Subscribe to updates.
-            async with async_timeout.timeout(5):
```
balloob, line 62: Without a timeout startup can hang forever if the node never responds?

agners: Matter SDK will timeout at one point... Maybe we should add a (higher) timeout at one point, but not sure what time that should be

PR #38, opened 2022-06-26
Add Matter Node Device

balloob:
Some Matter nodes are a bridge that expose multiple devices. This adds support for this.

See architecture drawing for latest architecture that includes Matter Node Device: https://docs.google.com/drawings/d/1hxhOXKquTUooWH6sSfCfw02fyfIOYL9MoxJf6EBUQ9o/edit


matter_server/client/model/mappable_device.py
```diff
@@ -0,0 +1,79 @@
...
+
+def mappable_devices_from_node(node: MatterNode) -> list[MappableDevice]:
+    """Split a node in mappable devices."""
+    if not node.bridge_device:
+        return [MappableNodeDevice(node)]
+
+    devices = []
+
+    for device in node.devices:
+        if device.device_type == device_types.BridgedDevice:
```
agners, line 25: Shouldn't that be `!=`?

agners: Hm, from what I can tell we exclude the `BridgedDevice` anyways when building `devices` list not?

agners (commented):
I guess there might be other cases where we need to split a node into multiple devices, but I guess that would need another `MappableDevice`. Needs to be implemented when it comes up.

matter_server/client/model/mappable_device.py
balloob: we exclude `Bridge`, not `BridgedDevice` . We use `BridgedDevice` to indicate which endpoints are their own devices.

balloob:
Marking as draft because I first want to merge entity description <> device mapping and then pick a better name for this too.

balloob:
Verified works.

tests/matter_experimental/test_mapping.py
```diff
@@ -0,0 +1,165 @@
...
+    async def connect(self):
+        self.server_info = Mock(compressedFabricId=MOCK_COMPR_FABRIC_ID)
+
+    async def listen(self, driver_ready):
+        self.driver = Driver(self)
+        driver_ready.set()
+        self.mock_client_disconnect = asyncio.Event()
+        await self.mock_client_disconnect.wait()
+
+    # async def async_send_command(
```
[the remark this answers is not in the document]
balloob: Removed 4f60fd92468d97fd79e7ed9809be08c223875938

PR #39, opened 2022-06-27
Merge entity description and device mapping

balloob:
This merges the device mapping and entity description classes, making it a single object to hold how a Matter device is mapped to an HA entity.

Haven't tested yet with real device.


custom_components/matter_experimental/binary_sensor.py
```diff
@@ -56,18 +60,28 @@ def _update_from_device(self) -> None:
...
+class MatterBinarySensorEntityDescription(
+    BinarySensorEntityDescription,
+    MatterEntityDescription,
+):
+    """Matter Binary Sensor entity description."""
+
+
+# You can't set default values on inherited data classes
+MatterSensorEntityDescriptionFactory = partial(
+    MatterBinarySensorEntityDescription, key=None, entity_cls=MatterBinarySensor
```
MartinHjelmare, line 73: Why do we not set a key? Shouldn't that be the device type?

balloob: Oh, yeah that's possible.

MartinHjelmare (approved):

PR #43, opened 2022-06-29
Extract device registry tests

balloob:
Extract device registry tests from the test_mapping test suite. This also generalizes the "set up node fixture" to be re-used across tests.


agners (approved):

tests/matter_experimental/common.py
```diff
@@ -0,0 +1,90 @@
...
+        self.server_info = Mock(compressedFabricId=MOCK_COMPR_FABRIC_ID)
+
+    async def listen(self, driver_ready):
+        self.driver = Driver(self)
+        driver_ready.set()
+        self.mock_client_disconnect = asyncio.Event()
+        await self.mock_client_disconnect.wait()
+
+
+@lru_cache(maxsize=None)
```
MartinHjelmare, line 43: Maybe use `cache` instead?

https://docs.python.org/3/library/functools.html#functools.cache

tests/matter_experimental/common.py
```diff
@@ -0,0 +1,90 @@
...
+from matter_server.vendor.chip.clusters.ObjectsVersion import CLUSTER_OBJECT_VERSION
+
+from tests.test_utils.mock_matter import get_mock_matter
+
+if TYPE_CHECKING:
+    from homeassistant.core import HomeAssistant
+
+
+MOCK_COMPR_FABRIC_ID = 1234
+LOGGER = logging.getLogger(__name__)
```
MartinHjelmare, line 26: LOGGER isn't used.

PR #46, opened 2022-07-06
Bump to CHIP version 5b603f3874 (2022-07-05 21:21:19 -0700)

agners:


agners:
Tested with actual devices, commissioning and controlling devices is still working :+1:

PR #78, opened 2022-09-20
Contact sensor test fixture

jculbert:
The change adds test fixture files for a contact sensor device. It adds one test to test_mapping which is passing on my system.

The content of cluster 1 was set from the connectedhomeip chef contact sensor example.

The content of cluster 0 was copied from the existing lighting-example-app.json. I found this to match what's in the chef contact sensor example. I made one minor change to the switch cluster in that "0, 1, 2" was added to attributeList.


balloob (approved):
Awesome, thank you!

PR #82, opened 2022-09-27
Add test fixture files for onoff switch

jculbert:
Added fixture files for on-off switch. The change adds one test which is passing on my system.

Note, the device type in the node file is set to 266 (0x010A), which maps to OnOffPlugUnit. This matches what is setup in matter_experimental/switch.py.


jculbert:
I just noticed that the serverList in the OnOffPlugUnit cluster is incorrect. I'll wait to see if there is another other feedback before fixing this.

tests/fixtures/nodes/onoff-switch-example-app.json
```diff
@@ -0,0 +1,771 @@
...
+        "attributeList": [0, 1, 2, 3, 65528, 65529, 65531, 65532, 65533],
+        "featureMap": 0,
+        "clusterRevision": 1,
+        "_type": "chip.clusters.Objects.SoftwareDiagnostics"
+      },
+      "ThreadNetworkDiagnostics": {
+        "TLVValue": {
+          "0": {
+            "TLVValue": null,
+            "Reason": "InteractionModelError: Failure (0x1)"
```
balloob, line 251: I guess we can drop all of `ThreadNetworkDiagnostics` since it all fails to decode anyway.

balloob (approved):
change looks good. Minor comment on included data.

balloob:
Thanks!

jculbert:
No problem.

PR #99, opened 2022-11-28
Refactor Matter server and client

marcelveldt:
Almost a complete rewrite, mainly because I needed to wrap my head around Matter and its models and because we changed the approach a bit.

- Server runs all business logic, client is transparent
- All data is stored with the server
- Same (clusters) models share between server and client
- Objects are serialized at transmission from server to client (or the other way around)
- Objects are deserialized (restructured back into correct class) from incoming json
- Server takes care of (re)interviewing nodes if needed

DRAFT, still some work in progress and last bugs to fix!


matter_server/common/models/error.py
```diff
@@ -0,0 +1,20 @@
...
+    """Error raised when interview of a device failed."""
+
+class NodeNotReady(MatterError):
+    """Error raised when performing action on node that has not been fully added."""
+
+class NodeNotExists(MatterError):
+    """Error raised when performing action on node that does not exist."""
+
+class VersionMismatch(MatterError):
+    """Issue raised when SDK version mismatches."""
```
MartinHjelmare, line 20: No newline at end of file.

matter_server/common/models/message.py
```diff
@@ -0,0 +1,73 @@
...
+
+from ..helpers.util import dataclass_from_dict
+from .events import EventType
+
+
+@dataclass
+class CommandMessage:
+    """Model for a Message holding a command from server to client or client to server."""
+
+    messageId: str
```
MartinHjelmare, line 16: Can we use snake_case attribute names instead?

matter_server/common/models/node.py
```diff
@@ -0,0 +1,133 @@
...
+    # attributes below will be auto derrived from the attributes
+    # root_device_type_instance: MatterDeviceTypeInstance[RootNode] | None = None
+    # aggregator_device_type_instance: MatterDeviceTypeInstance[Aggregator] | None = None
+    # device_type_instances: list[MatterDeviceTypeInstance] = field(default_factory=list)
+    # node_devices: list[AbstractMatterNodeDevice] = field(default_factory=list)
+
+    def parse_attributes(self, attributes: dict) -> None:
+        """Try to parse a MatterNode from AttributeCache (retrieved from Read)."""
+
+        device_type_instances: list[MatterDeviceTypeInstance] = []
```
MartinHjelmare, line 51: `device_type_instances` isn't used in this method.

matter_server/common/models/node.py
```diff
@@ -0,0 +1,133 @@
...
+                self.attributes[cluster.id] = cluster
+
+                # lookup device type for this cluster
+                device_type = ALL_TYPES.get(cluster.id)
+
+                if device_type is None:
+                    LOGGER.warning("Found unknown device type for Cluster %s", cluster)
+                    continue
+
+                instance = MatterDeviceTypeInstance(
```
MartinHjelmare, line 71: Where is this instance stored?

matter_server/common/models/node.py
```diff
@@ -0,0 +1,133 @@
...
+                if device_type is RootNode:
+                    self.root_device_type_instance = instance
+                elif device_type is Aggregator:
+                    self.aggregator_device_type_instance = instance
+                else:
+                    device_type_instances.append(instance)
+
+            self.device_type_instances = device_type_instances
+
+            if not hasattr(self, "root_device_type_instance"):
```
MartinHjelmare, line 111: It's not good to initialize attributes dynamically after init. Why do we not set this attribute in init?

matter_server/common/models/node.py
```diff
@@ -0,0 +1,133 @@
...
+
+from ..helpers.util import dataclass_to_dict
+from .device_type_instance import MatterDeviceTypeInstance
+from .node_device import (
+    AbstractMatterNodeDevice,
+    MatterBridgedNodeDevice,
+    MatterNodeDevice,
+)
+
+from chip.clusters import (
```
MartinHjelmare, line 20: Imports aren't sorted.

matter_server/common/models/node.py
```diff
@@ -0,0 +1,133 @@
...
+)
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class MatterNode:
+    """Matter node."""
+
+    nodeid: int
```
MartinHjelmare, line 36: Would be nice to consistently separate words with underscore.
```suggestion
    node_id: int
```

matter_server/server/client_handler.py
```diff
@@ -182,3 +133,84 @@ async def async_handle(self) -> web.WebSocketResponse:
...
+                    msg.messageId,
+                    ErrorCode.INVALID_COMMAND,
+                    f"Invalid command: {msg.command}",
+                )
+            )
+            self._logger.warning("Invalid command: %s", msg.command)
+            return
+
+        # schedule task to handle the command
+        self.server.loop.create_task(self._run_handler(handler, msg))
```
MartinHjelmare, line 156:
```suggestion
        asyncio.create_task(self._run_handler(handler, msg))
```

matter_server/server/client_handler.py
```diff
@@ -182,3 +133,84 @@ async def async_handle(self) -> web.WebSocketResponse:
...
+            self._logger.error(
+                "Client exceeded max pending messages: %s", MAX_PENDING_MSG
+            )
+
+            self._cancel()
+
+    def _cancel(self) -> None:
+        """Cancel the connection."""
+        if self._handle_task is not None:
+            self._handle_task.cancel()
```
MartinHjelmare, line 214: Do we need to await the task too to make sure it's cancelled, or is that not needed?

matter_server/server/device_controller.py
```diff
@@ -0,0 +1,391 @@
...
+        # Instantiate the underlying ChipDeviceController instance on the Fabric
+        self.chip_controller: ChipDeviceController = (
+            server.stack.fabric_admin.NewController()
+        )
+        # we keep the last events in memory so we can include them in the diagnostics dump
+        self.event_history: Deque[Attribute.EventReadResult] = deque(maxlen=25)
+        self._subscriptions: dict[int, Attribute.SubscriptionTransaction] = {}
+        self._nodes: dict[int, MatterNode] = {}
+        self._wifi_creds_set = False
+        self.logger.debug("CHIP Device Controller Initialized")
```
MartinHjelmare, line 86: It's nice to not create side effects when creating instances. Maybe we don't need this log?

matter_server/server/device_controller.py
```diff
@@ -0,0 +1,391 @@
...
+        # load nodes from persistent storage
+        nodes_data = self.server.storage.get(DATA_KEY_NODES, {})
+        for nodeid_str, node_dict in nodes_data.items():
+            nodeid = int(nodeid_str)
+            # TEMP !!!! TODO
+            # node_dict["attributes"] = {}
+            node = dataclass_from_dict(MatterNode, node_dict)
+            self._nodes[nodeid] = node
+            # make sure to start node subscriptions
+            await self._subscribe_node(nodeid)
```
MartinHjelmare, line 109:
```suggestion
        for node_id_str, node_dict in nodes_data.items():
            node_id = int(node_id_str)
            # TEMP !!!! TODO
            # node_dict["attributes"] = {}
            node = dataclass_from_dict(MatterNode, node_dict)
            self._nodes[node_id] = node
            # make sure to start node subscriptions
            await self._subscribe_node(node_id)
```

matter_server/server/device_controller.py
```diff
@@ -0,0 +1,391 @@
...
+        # create task to check for nodes that need any re(interviews)
+        self.server.loop.create_task(self._check_interviews())
```
MartinHjelmare, line 111:
```suggestion
        asyncio.create_task(self._check_interviews())
```

matter_server/server/device_controller.py
```diff
@@ -0,0 +1,391 @@
...
+    @api_command("device_controller.GetNodes")
+    def get_nodes(self) -> list[MatterNode]:
+        """Return all Nodes known to the server."""
+        return [x for x in self._nodes.values() if x is not None]
+
+    @api_command("device_controller.GetNode")
+    def get_node(self, nodeid: int) -> MatterNode:
+        """Return info of a single Node."""
+        node = self._nodes.get(nodeid)
+        assert node is not None, "Node does not exist or is not yet interviewed"
```
MartinHjelmare, line 132: This is just type annotation assertion or check for developers, right? If it can happen in production we shouldn't use `assert`.

matter_server/server/device_controller.py
```diff
@@ -0,0 +1,391 @@
...
+        # and/or have been interviewed more than 30 days ago.
+        for node in self._nodes.values():
+            if (
+                SCHEMA_VERSION > node.interview_version
+                or (datetime.utcnow() - node.last_interview).days > 30
+            ):
+                await self.interview_node(node.nodeid)
+        # reschedule self to run every hour
+        self.server.loop.call_later(
+            3600, self.server.loop.create_task, self._check_interviews()
```
MartinHjelmare, line 390:
```suggestion
            3600, asyncio.create_task, self._check_interviews()
```

matter_server/server/storage.py
```diff
@@ -0,0 +1,142 @@
...
+    from .server import MatterServer
+
+
+class StorageController:
+    """Controller that handles storage of persistent data."""
+
+    def __init__(self, server: "MatterServer") -> None:
+        """Initialize storage controller."""
+        self.server = server
+        self.logger = logging.getLogger(__name__)
```
MartinHjelmare, line 26: We could use a constant at the module level instead as we're just using the module logger here. We don't seem to need an instance attribute.

matter_server/server/storage.py
```diff
@@ -0,0 +1,142 @@
...
+        self._data: Dict[str, Any] = {}
+        self._timer_handle: asyncio.TimerHandle = None
```
MartinHjelmare, line 28:
```suggestion
        self._timer_handle: asyncio.TimerHandle | None = None
```

matter_server/server/storage.py
```diff
@@ -0,0 +1,142 @@
+"""Logic to handle storage of persistent data."""
+
```
MartinHjelmare, line 2:
```suggestion
from __future__ import annotations

```

matter_server/server/storage.py
```diff
@@ -0,0 +1,142 @@
...
+        """Load data from persistent storage."""
+        assert not self._data, "Already loaded"
+
+        def _load() -> dict:
+            for filename in self.filename, f"{self.filename}.backup":
+                try:
+                    _filename = os.path.join(self.server.storage_path, filename)
+                    with open(filename, "r") as _file:
+                        data = json_loads(_file.read())
+                        self.logger.debug(
```
MartinHjelmare, line 95: Maybe move the log and return to an `else:` block?

matter_server/server/device_controller.py
```diff
@@ -0,0 +1,391 @@
...
+        sub: Attribute.SubscriptionTransaction = await self.chip_controller.Read(
+            nodeid=nodeid, attributes="*", events="*", reportInterval=(0, 10)
+        )
+
+        def attribute_updated_callback(
+            path: Attribute.TypedAttributePath,
+            transaction: Attribute.SubscriptionTransaction,
+        ):
+            data = transaction.GetAttribute(path)
+            # value = {
```
MartinHjelmare, line 317: Commented code.

matter_server/common/helpers/util.py
```diff
@@ -0,0 +1,228 @@
...
+    return datetime.fromisoformat(datetimestr.replace("Z", "+00:00"))
+
+
+def parse_value(name: str, value: Any, value_type: Type, default: Any = MISSING):
+    """Try to parse a value from raw (json) data and type definitions."""
+
+    if isinstance(value_type, str):
+        value_type = eval(value_type)
+
+    # if issubclass(value_type, Cluster):
```
MartinHjelmare, line 104: Commented code.

matter_server/common/models/message.py
marcelveldt: sorry, leftover from the previous implementation, which in turn came from wave js node js implementation.
I'm in favor of adjusting it all to snake case

matter_server/common/helpers/util.py
```diff
@@ -0,0 +1,228 @@
...
+    # python 3.10
+    from types import NoneType
+except:  # noqa
+    # older python version
+    NoneType = type(None)
+
+CHIP_CLUSTERS_PKG_NAME = "home-assistant-chip-clusters"
+CHIP_CORE_PKG_NAME = "home-assistant-chip-core"
+
+# TODO: The below dataclass utils are shamelessly copied from aiohue
```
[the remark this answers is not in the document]
marcelveldt: yeah, I'm looking at that now because this still does not feel right. I have another serialization issue so I might end up just using dacite because all models are dataclasses.

matter_server/common/models/node.py
```diff
@@ -0,0 +1,133 @@
...
+    ClusterAttributeDescriptor,
+    ClusterEvent,
+    ClusterObjects,
+)
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class MatterNode:
```
MartinHjelmare, line 33: What's the difference between this `MatterNode` and the one in `common/models/device.py`?

matter_server/server/device_controller.py
marcelveldt: this is the part I'm still working on

matter_server/server/device_controller.py
```diff
@@ -0,0 +1,391 @@
...
+            await self._subscribe_node(nodeid)
+        # create task to check for nodes that need any re(interviews)
+        self.server.loop.create_task(self._check_interviews())
+        self.logger.debug("Started.")
+
+    async def stop(self) -> None:
+        """ "Handle logic on server stop."""
+        # unsubscribe all node subscriptions
+        for sub in self._subscriptions.values():
+            await self._call_sdk(sub.Shutdown)
```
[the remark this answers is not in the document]
marcelveldt: sequentially is better considered the sdk is a bit fragile sometimes for now.

matter_server/common/models/device.py
```diff
@@ -0,0 +1,37 @@
...
+class MatterAttribute(Attribute.AttributeCache):
+    """Representation of a Matter Attribute."""
+
+
+
+@dataclass
+class MatterNode:
+    """Representation of a Matter Node."""
+
+    nodeid: int
```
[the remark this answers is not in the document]
marcelveldt: well, the SDK has nodeid so I just used that to prevent confusion but no problem to adjust it.

pyproject.toml
```diff
@@ -21,8 +21,22 @@ classifiers = [
 ]
 dependencies = [
   "aiohttp",
+  "aiorun",
```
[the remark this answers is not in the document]
MartinHjelmare: Ok.

matter_server/common/models/node.py
marcelveldt: WIP, I was working on that now and removed some code by accident, thanks for the reminder ;-)

matter_server/common/models/node.py
marcelveldt: This is also just copied from the previous implementation and I did not have a chance to fix it

matter_server/server/client_handler.py
marcelveldt: nah, in here it is not that important I guess

matter_server/server/device_controller.py
marcelveldt: Yeah, this log can be moved to the start function

matter_server/server/device_controller.py
marcelveldt: Do we really want to change it to "node_id" if the SDK calls it "nodeid" ? I see this biting us some time ;-) 
From the other hand we should create our own api we feel comfortable with and not mimic another...

Just as with the camel_case vs snake_case we discussed before, I'm in favor of changing it to our own liking

matter_server/server/device_controller.py
marcelveldt: yeah, used it while testing. To be removed later

matter_server/server/storage.py
marcelveldt: look like some leftover debug

matter_server/common/models/node.py
marcelveldt: see above, WIP

matter_server/common/models/node.py
balloob: Can we do the cleanup in a follow-up PR and instead focus on merging + getting it working in HA?

matter_server/server/device_controller.py
MartinHjelmare: I'd prefer always calling it `node_id`. I don't think we'll get confused about what it is.

matter_server/common/helpers/util.py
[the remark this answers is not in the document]
marcelveldt: too many weird value types so using dacite turned out to be a nightmare. The custom code is now (finally) working well for everything I have tested so far and its now able to dynamically serialize to/from the data class models of the clusters package, including the weird stuff with like actually class types as dict keys.

We can improve and cleanup later but for now it works. Spent too many day trying to get it to work with existing helper packages.

matter_server/common/models/node.py
marcelveldt: left out for now. will fix in follow up while working on core part as I'm not 100% how to do the discovery properly

matter_server/common/helpers/util.py
```diff
@@ -101,6 +103,12 @@ def parse_value(name: str, value: Any, value_type: Type, default: Any = MISSING)
     if isinstance(value_type, str):
         value_type = eval(value_type)
 
+    if isinstance(value, dict):
+        if hasattr(value_type, "from_dict"):
+            return value_type.from_dict(value)
+        if hasattr(value_type, "FromDict"):
+            return value_type.from_dict(value)
```
balloob: Shouldn't this be? Also, which models have these?
```suggestion
        if hasattr(value_type, "FromDict"):
            return value_type.FromDict(value)
```

marcelveldt: it's not in use actually

balloob: Then remove it in a future PR

matter_server/client/matter.py
```diff
@@ -1,250 +0,0 @@
-"""Matter base class."""
```
balloob, line 1: Discussed offline with Marcel. We're going to re-instate this file again to split again the Matter from WebSocket logic.

marcelveldt:
as greed, we're going to merge this one and fix remaining issues and improvements later

PR #112, opened 2022-12-01
Fix issues with parsing models on Python 3.9

marcelveldt:
Union types need to be written the oldskool way (even with annotations from future), otherwise our dynamic model creation does not work on python 3,9 causing import errors.

### DIFF
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -35,7 +35,7 @@ try:
 except:  # noqa
     # older python version
     NoneType = type(None)
-    UnionType = type(int | str)
+    UnionType = type(Union)
 
 CHIP_CLUSTERS_PKG_NAME = "home-assistant-chip-clusters"
 CHIP_CORE_PKG_NAME = "home-assistant-chip-core"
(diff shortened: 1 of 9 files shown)


matter_server/common/helpers/util.py
```diff
@@ -35,7 +35,7 @@
 except:  # noqa
     # older python version
     NoneType = type(None)
-    UnionType = type(int | str)
```
balloob, line 38: The reason this doesn't work with `from __future__` is because it is not written inside an annotation.

matter_server/common/helpers/util.py
```diff
@@ -87,15 +87,20 @@ def parse_utc_timestamp(datetimestr: str):
     return datetime.fromisoformat(datetimestr.replace("Z", "+00:00"))
 
 
-def parse_value(name: str, value: Any, value_type: Type | str, default: Any = MISSING):
+def parse_value(
+    name: str, value: Any, value_type: Union[Type, str], default: Any = MISSING
```
balloob, line 91: this is inside an annotation so would have been fine.

marcelveldt: Yeah, unrelated. The issue was the parsing of the actual annotations in this function itself.

matter_server/common/helpers/util.py
marcelveldt: Yeah, missed that while in the rush

PR #118, opened 2022-12-01
Continue test steps even on error

agners:


.github/workflows/test.yml
```diff
@@ -12,6 +12,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    continue-on-error: true
     strategy:
```
ludeeus, line 16:
```suggestion
    strategy:
      fail-fast: false
```
Should do this as well so they do not abort eachother

ludeeus (approved):

PR #120, opened 2022-12-01
Use all instead of disabling flake8

marcelveldt:
@MartinHjelmare


MartinHjelmare (approved):
Thanks!

matter_server/server/__init__.py
```diff
@@ -1,3 +1,5 @@
 """Implementation of a Websocket-based Matter proxy (using CHIP SDK)."""
 
 from .server import MatterServer  # noqa
```
MartinHjelmare, line 3:
```suggestion
from .server import MatterServer
```

PR #136, opened 2022-12-14
Add pylint

MartinHjelmare:
- Add pylint to test requirements and activate pylint in CI.


MartinHjelmare:
We need a wheel for `pygobject`. Either a ready wheel or be able to build it during CI.

.github/workflows/test.yml
```diff
@@ -30,6 +30,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
+          sudo apt-get update
+          sudo apt-get install python3-gi
```
ludeeus, line 34:
```suggestion
          sudo apt-get install -y python3-gi
```

PR #141, opened 2022-12-17
Type package

MartinHjelmare:
- Move towards complete type annotations.
- Add py.typed.
- Add mypy during CI.

### DIFF
--- a/matter_server/common/helpers/api.py
+++ b/matter_server/common/helpers/api.py
@@ -1,7 +1,9 @@
 """Several helpers for the WebSockets API."""
+from __future__ import annotations
+
 from dataclasses import MISSING, dataclass
 import inspect
-from typing import Any, Callable, Coroutine, Optional, TypeVar
+from typing import Any, Callable, Coroutine, TypeVar
 
 from matter_server.common.helpers.util import parse_value
 
@@ -45,7 +47,7 @@ def get_typed_signature(call: Callable) -> inspect.Signature:
 
 
 def parse_arguments(
-    func_sig: inspect.Signature, args: Optional[dict], strict: bool = False
+    func_sig: inspect.Signature, args: dict | None, strict: bool = False
 ) -> dict[str, Any]:
     """Parse (and convert) incoming arguments to correct types."""
     if args is None:
--- a/matter_server/common/helpers/api.py
+++ b/matter_server/common/helpers/api.py
@@ -1,9 +1,11 @@
 """Several helpers for the WebSockets API."""
 from dataclasses import MISSING, dataclass
 import inspect
-from typing import Any, Callable, Coroutine, Optional
+from typing import Any, Callable, Coroutine, Optional, TypeVar
 
 from matter_server.common.helpers.util import parse_value
+
+_F = TypeVar("_F", bound=Callable[..., Any])
 
 
 @dataclass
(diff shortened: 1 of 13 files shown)


MartinHjelmare (commented):
Not sure what to do here:
https://github.com/home-assistant-libs/python-matter-server/blob/6bf1b5990fd57cba67a3ebc2631553057bf3face/matter_server/server/client_handler.py#L180

The `args` attribute of `CommandMessage` is optional. `parse_arguments` doesn't handle an optional parameter.

MartinHjelmare:
I think I first need to fix CI so we don't lint on both Python 3.9 and Python 3.10.

matter_server/common/helpers/util.py
```diff
@@ -191,26 +194,21 @@ def parse_value(
...
 
-def dataclass_from_dict(cls: Optional[dataclass], dict_obj: dict, strict=False):
+def dataclass_from_dict(cls: type[_T], dict_obj: dict, strict: bool = False) -> _T:
     """
     Create (instance of) a dataclass by providing a dict with values.
 
     Including support for nested structures and common type conversions.
     If strict mode enabled, any additional keys in the provided dict will result in a KeyError.
     """
-    if "_type" in dict_obj:
```
marcelveldt, line 209: why did you remove this ?

marcelveldt: nvm, got my answer :-)

marcelveldt:
> I think I first need to fix CI so we don't lint on both Python 3.9 and Python 3.10.

I think we do need to lint on python 3.9 and python3.10 because most issues so far have been py 3.9 compatibility issues due to difference behavior in annotations between python 3.9 and 3.10. I would really like to drop 3.9 to get rid of some workarounds.

marcelveldt:
> Not sure what to do here:
> 
> https://github.com/home-assistant-libs/python-matter-server/blob/6bf1b5990fd57cba67a3ebc2631553057bf3face/matter_server/server/client_handler.py#L180
> 
> The `args` attribute of `CommandMessage` is optional. `parse_arguments` doesn't handle an optional parameter.

It should handle it already but the typing needs to be fixed. parse_arguments looks at the function signature to detect what arguments are required (and their typings) and compares that to the incoming data. We should default to an empty dict if args is None.

MartinHjelmare:
It's easier if we just lint on min Python version and run tests on all versions to verify behavior.

matter_server/common/helpers/api.py
```diff
@@ -1,5 +1,5 @@
 """Several helpers for the websockets API."""
-
+from __future__ import annotations
```
marcelveldt, line 2: look out for side effects here.. Does the utility still work with this import on python 3.9 ?
we are using the inspect module to grab annotations for functions, the big difference between py3.9 and py3.10 is if the output is strings or actual classes/types. I remember the future import messing with that..

MartinHjelmare: After we merge https://github.com/home-assistant-libs/python-matter-server/pull/147 I think we can go ahead with more confidence here if tests pass.

marcelveldt: I'm currently testing this branch to look for any side effects

marcelveldt: everything still works perfectly fine, tested with py3.10 and 3.9 and everything still working smooth, couldn't detect any side efefcts

marcelveldt (approved):

PR #142, opened 2022-12-18
Split lint and test in CI

MartinHjelmare:
- Split lint and test in two different jobs.
- The lint job should only run on the minimum Python version. The test job should run on all supported minor Python versions.
- Disable the test job for now, until the tests are fixed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.2.0
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python
         uses: actions/setup-python@v4.3.1
         with:
           python-version: "3.9"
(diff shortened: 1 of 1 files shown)


agners (approved):

.github/workflows/test.yml
```diff
@@ -33,18 +59,10 @@ jobs:
...
-      #   run: python3 -m scripts.show_stored_node tests/fixtures/nodes/lighting-example-app.json
-      # - name: test dump fixer script
-      #   run: python3 -m scripts.dump_fixer tests/fixtures/nodes/lighting-example-app.json
-      # - name: Pytest
-      #   run: pytest tests
+          pip install -e .[server] -r requirements-test.txt
+      - name: test show stored node script
+        run: python3 -m scripts.show_stored_node tests/fixtures/nodes/lighting-example-app.json
+      - name: test dump fixer script
+        run: python3 -m scripts.dump_fixer tests/fixtures/nodes/lighting-example-app.json
```
marcelveldt, line 66: these should not be enabled again as they are not adjusted to the code changes

MartinHjelmare: Yeah, the whole job is disabled.

PR #143, opened 2022-12-19
Fix Commissioning

marcelveldt:
With latest changes of the CHIP SDK the commissioning now works as intended and we can remove the workaround.
Note that this does not yet fix the attestation error for devices without valid certificate.

Depend on chip wheels 2022.12.0 (which is bumped in this PR)

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -65,7 +65,9 @@ class MatterDeviceController:
 
     async def start(self) -> None:
         """Async initialize of controller."""
-        self.compressed_fabric_id = await self._call_sdk(self.chip_controller.GetCompressedFabricId)
+        self.compressed_fabric_id = await self._call_sdk(
+            self.chip_controller.GetCompressedFabricId
+        )
         # load nodes from persistent storage
         nodes_data = self.server.storage.get(DATA_KEY_NODES, {})
         for node_id_str, node_dict in nodes_data.items():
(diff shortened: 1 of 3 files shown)


matter_server/server/device_controller.py
```diff
@@ -56,19 +55,18 @@ def __init__(
         self._nodes: dict[int, MatterNode] = {}
         self.wifi_credentials_set: bool = False
         self.thread_credentials_set: bool = False
+        self.compressed_fabric_id: int | None = None
```
MartinHjelmare, line 58: I think we need to adapt the `get_info` method in the server to check for optional value after this change.

matter_server/server/device_controller.py
```diff
@@ -114,22 +112,15 @@ async def commission_with_code(self, code: str) -> MatterNode:
...
-        await self._call_sdk(
+        success = await self._call_sdk(
             self.chip_controller.CommissionWithCode,
             setupPayload=code,
             nodeid=node_id,
         )
-        await asyncio.sleep(60)
+        if not success:
+            raise NodeCommissionFailed(
+                f"Commission with code failed for node {node_id}"
```
MartinHjelmare, line 122: Side note: There's no way to get more context why the commissioning failed?

marcelveldt: No, unfortunately not. The error is not exposed. Good news is that this is changed upstream in the sdk code so in matter 1.1 release we will have more context

matter_server/server/device_controller.py
marcelveldt: Well that is a theoretical discussion because get-info can not be called without the stack being initialized (and thius the compressed_fabric_id known). For now I've just added an assertion there.

MartinHjelmare: Yes, that's enough.

MartinHjelmare (approved):

PR #147, opened 2022-12-21
Add server test foundation

MartinHjelmare:
- Add server tests.
- Activate server tests in CI.
- Fix `parse_arguments` optional args.

### DIFF
--- a/matter_server/common/helpers/api.py
+++ b/matter_server/common/helpers/api.py
@@ -1,9 +1,7 @@
 """Several helpers for the WebSockets API."""
-
-
 from dataclasses import MISSING, dataclass
 import inspect
-from typing import Any, Callable, Coroutine
+from typing import Any, Callable, Coroutine, Optional
 
 from matter_server.common.helpers.util import parse_value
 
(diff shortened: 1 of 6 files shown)


matter_server/common/helpers/api.py
```diff
@@ -44,8 +42,12 @@ def get_typed_signature(call: Callable) -> inspect.Signature:
     return signature
 
 
-def parse_arguments(func_sig: inspect.Signature, args: dict, strict: bool = False):
+def parse_arguments(
+    func_sig: inspect.Signature, args: Optional[dict], strict: bool = False
```
MartinHjelmare, line 46: I included this fix to make the tests pass when passing `None` for `args`, as we also do sometimes in the server.

I can break this commit out if wanted.

marcelveldt: its fine like this, keep it this way. In the long run I want to have a second look at that whole argument parsing thingie but not for today

marcelveldt (approved):
nice!

PR #160, opened 2023-01-08
Fix persistency were controller failed loading commissioned devices

ArturoGuerra:
`compressed_fabric_id` isn't set until the device controller is initialized, and we need the fabric id for loading from the storage path.

### DIFF
--- a/matter_server/server/server.py
+++ b/matter_server/server/server.py
@@ -85,8 +85,11 @@ class MatterServer:
                 "CHIP Core version does not match CHIP Clusters version."
             )
         self.loop = asyncio.get_running_loop()
-        await self.storage.start()
+
         await self.device_controller.start()
+        await self.storage.start()
+        await self.device_controller.load_nodes()
+
         mount_websocket(self, "/ws")
         self.app.router.add_route("GET", "/", self._handle_info)
         self._runner = web.AppRunner(self.app, access_log=None)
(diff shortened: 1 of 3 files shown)


matter_server/server/server.py
```diff
@@ -85,6 +85,7 @@ async def start(self) -> None:
                 "CHIP Core version does not match CHIP Clusters version."
             )
         self.loop = asyncio.get_running_loop()
+        await self.device_controller.initialize()
```
MartinHjelmare, line 88: The issue is that `compressed_fabric_id` isn't set until the device controller is initialized, and we need the fabric id for loading from the storage path, right?

Dockerfile.dev
```diff
@@ -27,9 +27,9 @@ RUN \
     git \
     && git clone --depth 1 -b master \
     https://github.com/project-chip/connectedhomeip \
-    && cp -r connectedhomeip/credentials /root/credentials \
-    && mv /root/credentials/production/paa-root-certs/* \
-          /root/credentials/development/paa-root-certs/ \
+    && cp -r connectedhomeip/credentials /app/credentials \
```
MartinHjelmare, line 30: Please rebase on latest main branch. These changes are already merged.

matter_server/server/server.py
ArturoGuerra: yep

MartinHjelmare (approved):
Thanks!

PR #166, opened 2023-01-12
split websocket connection to own file

marcelveldt:


matter_server/client/client.py
```diff
@@ -251,78 +236,30 @@ async def send_command_no_wait(
...
-        )
+        """Connect to the Matter Server (over Websockets)."""
+        self.loop = asyncio.get_running_loop()
+        if self.connection.connected:
+            # already connected
+            return
+        await self.connection.connect()
+        # connect will raise when connecting failed,
+        # otherwise signal connected event
+        self._signal_event(EventType.CONNECTED)
```
MartinHjelmare, line 250: This is a new feature, right? I think we should wait with those.

matter_server/client/client.py
```diff
@@ -53,16 +41,18 @@ class MatterClient:
 
     def __init__(self, ws_server_url: str, aiohttp_session: ClientSession):
         """Initialize the Client class."""
-        self.ws_server_url = ws_server_url
+        self.loop: asyncio.AbstractEventLoop | None = None
```
MartinHjelmare, line 44: I don't think the loop should be a feature for library users. Can we keep it protected?

matter_server/client/client.py
```diff
@@ -53,16 +41,18 @@ class MatterClient:
 
     def __init__(self, ws_server_url: str, aiohttp_session: ClientSession):
         """Initialize the Client class."""
-        self.ws_server_url = ws_server_url
```
MartinHjelmare, line 56: Removing the server url is a breaking change. I think it's a good change, but we should call it out in the PR description and label the PR as breaking, in that case.

MartinHjelmare (commented):
There a lint failures.

matter_server/client/client.py
marcelveldt: well, strictly takes yes but non breaking/very small. I'll take it out

matter_server/client/client.py
marcelveldt: We may want it later in the sub controllers ?

matter_server/client/client.py
marcelveldt: Are we using it in the HA integration ? If so I can expose it as a property.
If not, we should not care at this point imo.

matter_server/client/client.py
```diff
@@ -53,16 +41,18 @@ class MatterClient:
...
+        self.loop: asyncio.AbstractEventLoop | None = None
+        self.connection = MatterClientConnection(ws_server_url, aiohttp_session)
         self.logger = logging.getLogger(__package__)
-        self.aiohttp_session = aiohttp_session
```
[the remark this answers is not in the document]
marcelveldt: see above, I'm not aware of any other consumers than HA so if we're not relying on it there we can afford these small changes at this stage of the project. Just my 2 cents...

matter_server/client/client.py
```diff
@@ -104,11 +94,6 @@ def unsubscribe() -> None:
 
         return unsubscribe
 
-    @property
-    def connected(self) -> bool:
```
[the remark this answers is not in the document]
marcelveldt: same comment, my proposal is to expose the properties that we actually use in HA, if not, skip them.

matter_server/client/client.py
MartinHjelmare: I'd rather get it there as needed than make it non protected attribute.

matter_server/client/client.py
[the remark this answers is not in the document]
marcelveldt: and yeah, note the breaking change in the PR description if somebody is actually using our lib now

matter_server/client/client.py
MartinHjelmare: I'm not sure. Regardless we should document the changes in the release notes. Labeling it here makes that easier.

matter_server/client/connection.py
```diff
@@ -0,0 +1,166 @@
...
+    """Manage a Matter server over WebSockets."""
+
+    def __init__(
+        self,
+        ws_server_url: str,
+        aiohttp_session: ClientSession,
+    ):
+        """Initialize the Client class."""
+        self.ws_server_url = ws_server_url
+        self.aiohttp_session = aiohttp_session
```
MartinHjelmare, line 42: I'm wondering if we should make the session protected. It's not meant for anything else to use, right?

marcelveldt: yeah, good idea, I have no clue why you would want to make it public

marcelveldt: adjusted it

PR #167, opened 2023-01-12
Expose diagnostics in the client

marcelveldt:


MartinHjelmare (approved):
Thanks!

Can we make a new release now, with this addition?

PR #170, opened 2023-01-15
Set network mode host and add default USERDIR in docker-compose.yml

ArturoGuerra:
Running in host mode is required for mDNS discovery which is needed when commissioning devices that are already on the network like the hue bridge.


MartinHjelmare (approved):
Thanks!

Issue #171, opened 2023-01-15
pypi.org page is lacking updated information


guyluz11:
There are several missing and old information on readme and pypi.org  pages

First of all the [pypi.org](https://pypi.org/project/python-matter-server) page is missing a link to this repo

![image](https://user-images.githubusercontent.com/9304740/212568649-d6e5b9cc-2f8d-4246-a73d-443b9af95ae7.png)

Second there are several places in the readme that says the old name `chip` and didn't got updated to the new name `matter`.

Third the command `python -m matter_server.server` does not work for me, maybe I am missing something.
Getting the error 
```python
python -m matter_server.server
/usr/bin/python: No module named matter_server
```

kjagiello:
> Third the command `python -m matter_server.server` does not work for me, maybe I am missing something.
Getting the error

Are you sure you have installed this package? I set it up yesterday and can successfully boot up the server using that command.

marcelveldt:
The upstream sdk is still named project chip, hence the name

guyluz11:
> > Third the command `python -m matter_server.server` does not work for me, maybe I am missing something.
> > Getting the error
> 
> Are you sure you have installed this package? I set it up yesterday and can successfully boot up the server using that command.

My bad, I run it on python (python2) and not on python3 so it got confused.

I am getting a different error now 


 ```python
INFO Initializing CHIP/Matter Controller Stack...
[1674064152.017892][8595:8595] CHIP:CTL: Setting attestation nonce to random value
[1674064152.019712][8595:8595] CHIP:CTL: Setting CSR nonce to random value
[1674064152.024594][8595:8595] CHIP:DL: ChipLinuxStorage::Init: Using KVS config file: /tmp/chip_kvs
[1674064152.024714][8595:8595] CHIP:DL: writing settings to file (/tmp/chip_kvs-FjN1PO)
[1674064152.024766][8595:8595] CHIP:DL: renamed tmp file to file (/tmp/chip_kvs)
[1674064152.024836][8595:8595] CHIP:DL: ChipLinuxStorage::Init: Using KVS config file: /data/chip_factory.ini
[1674064152.024849][8595:8595] CHIP:DL: failed to open file (/data/chip_factory.ini-h1JPBd) for writing
[1674064152.024858][8595:8595] CHIP:DL: Configuration Manager initialization failed: ../src/platform/Linux/CHIPLinuxStorageIni.cpp:122: CHIP Error 0x000000AD: Open file failed
[...]

PR #172, opened 2023-01-15
Readme changed Chip to Matter

guyluz11:
Small readme fix that is part of #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Python Matter Server
 
-This project implements a Matter Controller Server over WebSockets using the [official CHIP SDK](https://github.com/project-chip/connectedhomeip) as a base and provides both a server and client implementation.
+This project implements a Matter Controller Server over WebSockets using the [official Matter SDK](https://github.com/project-chip/connectedhomeip) as a base and provides both a server and client implementation.
 
 The goal of this project is primary to have Matter support in Home Assistant but its universal approach makes it suitable to be used in other projects too.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Python Matter Server
 
-This project implements a Matter Controller Server over WebSockets using the [official CHIP SDK](https://github.com/project-chip/connectedhomeip) as a base and provides both a server and client implementation.
+This project implements a Matter Controller Server over WebSockets using the [official Matter SDK](https://github.com/project-chip/connectedhomeip) as a base and provides both a server and client implementation.
```
marcelveldt, line 3:
```suggestion
This project implements a Matter Controller Server over WebSockets using the [official Matter (formerly CHIP) SDK](https://github.com/project-chip/connectedhomeip) as a base and provides both a server and client implementation.
```

README.md
```diff
@@ -15,7 +15,7 @@ NOTE: Both Matter and this implementation are in early (v1) state and features a
...
-The client library has a dependency on the chip clusters package which contains all (Cluster) models and this package is os/platform independent. The server library depends on the CHIP Core SDK which is architecture and OS specific. We build (and publish) wheels for Linux (amd64 and aarch64) to pypi but for other platforms (like Macos) you will need to build those wheels yourself using the exact same version of the SDK as we use for the clusters package. Take a look at our build script for directions: https://github.com/home-assistant-libs/chip-wheels/blob/main/.github/workflows/build.yaml
+The client library has a dependency on the matter clusters package which contains all (Cluster) models and this package is os/platform independent. The server library depends on the Matter Core SDK which is architecture and OS specific. We build (and publish) wheels for Linux (amd64 and aarch64) to pypi but for other platforms (like Macos) you will need to build those wheels yourself using the exact same version of the SDK as we use for the clusters package. Take a look at our build script for directions: https://github.com/home-assistant-libs/chip-wheels/blob/main/.github/workflows/build.yaml
```
marcelveldt, line 18:
```suggestion
The client library has a dependency on the chip/matter clusters package which contains all (Cluster) models and this package is os/platform independent. The server library depends on the Matter Core SDK (still named CHIP) which is architecture and OS specific. We build (and publish) wheels for Linux (amd64 and aarch64) to pypi but for other platforms (like Macos) you will need to build those wheels yourself using the exact same version of the SDK as we use for the clusters package. Take a look at our build script for directions: https://github.com/home-assistant-libs/chip-wheels/blob/main/.github/workflows/build.yaml
```

marcelveldt:
Well, the SDK is still named CHIp and the wheels are being published as CHIP but I agree we should try to avoid the CHIP name. I've proposed some changes to seek some middle ground where we explain that CHIP is the previous name of matter.

guyluz11:
Let's wait for results from https://github.com/project-chip/connectedhomeip/issues/24436

PR #189, opened 2023-02-03
Add support for ValueDecodeFailure

ArturoGuerra:
Handles bad matter devices that have cluster attributes with the wrong values and are turned into `ValueDecodeFailure` types by the chip library.

Resolves https://github.com/home-assistant-libs/python-matter-server/issues/187

Hue Bridge is an example of a device in this case an ExtendedColorLight with attributes that can't be parsed by the chip controller
Hue Bridge Example

### DIFF
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -23,6 +23,7 @@ import chip.clusters  # noqa: F401
 from chip.clusters import Objects  # noqa: F401
 from chip.clusters.Objects import *  # noqa: F401 F403
 from chip.clusters.Types import Nullable, NullValue  # noqa: F401
+from chip.clusters.Attribute import ValueDecodeFailure # noqa: F401
 from chip.tlv import float32, uint
 
 from ..models.events import *  # noqa: F401 F403
@@ -46,7 +47,6 @@ except:  # noqa
     # older python version
     NoneType = type(None)
     UnionType = type(Union)
-
 
 _T = TypeVar("_T")
 
@@ -119,12 +119,14 @@ def parse_value(
             value_type = eval(value_type)
         except TypeError:
             pass
-
+    
     elif isinstance(value, dict):
         if hasattr(value_type, "from_dict"):
             return value_type.from_dict(value)
         if hasattr(value_type, "FromDict"):
             return value_type.FromDict(value)
+    
+    #logging.getLogger(__name__).warn("parse_value: %s %s %s %s", name, value, type(value), value_type)
 
     if value is None and not isinstance(default, type(MISSING)):
         return default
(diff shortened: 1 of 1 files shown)


matter_server/common/helpers/util.py
```diff
@@ -159,6 +160,10 @@ def parse_value(
                 return parse_value(name, value, sub_arg_type)
             except (KeyError, TypeError, ValueError):
                 pass
+        # Before we give up, check if we have a ValueDecodeFailure
+        if isinstance(value, dict) and "TLVValue" in value and "Reason" in value:
+            return ValueDecodeFailure(value["TLVValue"], value["Reason"])
```
MartinHjelmare, line 165: Where and how will this instance be handled downstream?

ArturoGuerra: Well currently it's not handled at all, it just gracefully replaces the correct field with that which is the same behavior the chip cli has, it could probably be set to some default to make sure it's the correct type instead of returning `ValueDecodeFailed`

ArturoGuerra: idk which approach you guys would prefer for this, I do think handling it gracefully is important as devices having unparsable values is a real possibility like the hue bridge but at the same time getting errors is also important so maybe adding a strict mode that errors out and a normal mode that just gracefully returns is a good solution?

MartinHjelmare: I think we need to think about and discuss how we want to handle this where we use the util, both here in this library and when library users use the util, eg Home Assistant.

What are the different usages and what should happen when we encounter this case?

ArturoGuerra: Well the only usage I've seen for this is to get specific attributes from a cluster without having to subscribe to events, some examples would be getting the min/max brightness of a light, getting the min/max values for color, and in my specific case, it was checking if the cluster has a specific feature enabled using the featureMap attribute, so it's mostly about getting static data from a device.

As to handling this from what I've seen on the only other matter controller I have working which is apple home they just ignore it completely as the device works, now this might be because looking at the matter spec this is only needed if a specific feature is enabled in this case color loop which hue has disabled, with that being said they should still be returning parsable data for this attribute, in this case, it should be a null. So I think we should just return a `ValueDecodeFailure` if that's what we get from the matter API and we should also add a function to get specific attributes from a cluster which should error out if the value is of an incorrect type, something that should be simple enough to add since the `MatterNode` class already has a function for fetching specific attributes it just doesn't check if the type is correct.

balloob:
Marking as draft pending https://github.com/orgs/home-assistant-libs/projects/1?pane=issue&itemId=16855348

marcelveldt:
I have seen this one too while working on the bridge support, its pretty common. 
For now I've filtered out the ones I saw but we need to fix this both upstream and in the parser logic at some point.

PR #191, opened 2023-02-04
Add helper method get_attribute to MatterDeviceTypeInstance

ArturoGuerra:
This method makes it easier to get a `MatterAttribute` from a `MatterDeviceTypeInstance`

### DIFF
--- a/matter_server/common/models/device_type_instance.py
+++ b/matter_server/common/models/device_type_instance.py
@@ -64,15 +64,8 @@ class MatterDeviceTypeInstance(Generic[_DEVICE_TYPE_T]):
     
     def get_attribute(self, cluster: type[all_clusters.Cluster], attribute: Union[str, int, type]) -> MatterAttribute:
         """Gets an attribute from a cluster for this device type instance."""
-        if not self.has_cluster(cluster):
-            raise ValueError(f"Device does not have cluster {cluster}")
-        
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
What's the current alternative to the new helper?

ArturoGuerra:
It would be something like this `instance.node.get_attribute(instance.endpoint, Clusters.LevelControl, "minLevel")` and with this it would just be `instance.get_attribute(Clusters.LevelControl, "minLevel")`

ArturoGuerra:
It's related to the last part of the comment here https://github.com/home-assistant-libs/python-matter-server/pull/189#discussion_r1096370465

balloob:
I suggest that we don't merge this right now. Marcel is planning on re-doing how data is stored both on the server and the client side as tracked here: https://github.com/orgs/home-assistant-libs/projects/1?pane=issue&itemId=16855348

ArturoGuerra:
in that case, we should also hold off on this pull request https://github.com/home-assistant-libs/python-matter-server/pull/189

marcelveldt:
I was thinking of fixing the bridge support and then look at the datamodel but I guess we need to do the datamodel first and then fix the bridges

marcelveldt:
This one can be closed now that its all handled at endpoint level

PR #195, opened 2023-02-11
Additional commands for the documentation

flatsiedatsie:
and slight change of some existing commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ When the start_listening command is issued, the server will dump all existing no
 
 
 **Send a command**
-This is a little bit more involved. Here is an example of turning on a switch.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,6 +148,39 @@ When the start_listening command is issued, the server will dump all existing no
   }
 ```
 
+
+**Send a command**
+This is a little bit more involved. Here is an example of turning on a switch.
```
marcelveldt, line 153:
```suggestion
Because we use the datamodels of the Matter SDK, this is a little bit more involved. Here is an example of turning on a switch.
```

marcelveldt (approved):
Great, thanks!

PR #197, opened 2023-02-13
Add basic pre-commit checks

ArturoGuerra:
Adds some basic pre-commit checks and modified required files to pass all added checks:
black
codespell
flake8
isort
mypy


ArturoGuerra:
Well thats unexpected

ArturoGuerra:
Okay it appears this is due to the fact the current CI pipeline is still using python3.9 instead of python3.10

scripts/run-in-env.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
```
MartinHjelmare, line 1: Do we really need this script? I'd assume that the contributor has the environment active when committing.

ArturoGuerra: Probably not only reason I added it was because the HA core repo has it

MartinHjelmare: Yeah, I think it was added there when some contributors were running GUI tools that didn't activate the environment automatically. I don't think we need it here as we'll have a smaller contributor community. We can update later if we see a need.

ArturoGuerra: Kk I'll remove in the morning 😃

ArturoGuerra:
The only thing missing rn is mypy but we can add that once support for python3.9 is removed

PR #207, opened 2023-02-15
Change schema version logic

marcelveldt:
Make the schema version checks less strict so we have some more freedom of updating the addon/core independently.

- Remove strict check of sdk version alltogether. If there are any breaking changes, we can bump the schema version accordingly. NOTE to myself is to include the sdk version in the diagnostics when I'm reworking that.

- Use schema version constant (integer value) which we can bump if there are any breaking changes or new features.

- Define minimal supported schema version in the server itself, we should just keep things backwards compatible.
- Log warning if schema version does not match
- Keep logic in place to tie a feature to a schema version

### DIFF
--- a/matter_server/client/connection.py
+++ b/matter_server/client/connection.py
@@ -9,7 +9,7 @@ from typing import Any, Callable, Dict, Final, cast
 from aiohttp import ClientSession, ClientWebSocketResponse, WSMsgType, client_exceptions
 
 from ..common.helpers.json import json_dumps, json_loads
-from ..common.helpers.util import chip_clusters_version, parse_message
+from ..common.helpers.util import parse_message
 from ..common.models.events import EventType
 from ..common.models.message import CommandMessage, MessageType, ServerInfoMessage
 from ..common.models.node import MatterNode
--- a/matter_server/client/connection.py
+++ b/matter_server/client/connection.py
@@ -14,7 +14,7 @@ from ..common.models.events import EventType
 from ..common.models.message import CommandMessage, MessageType, ServerInfoMessage
 from ..common.models.node import MatterNode
 from ..common.models.server_information import ServerInfo
-from .const import MIN_SCHEMA_VERSION
+from ..common.const import MIN_SCHEMA_VERSION, MAX_SCHEMA_VERSION, SCHEMA_VERSION
 from .exceptions import (
     CannotConnect,
     ConnectionClosed,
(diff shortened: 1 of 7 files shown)


matter_server/client/connection.py
```diff
@@ -92,14 +84,20 @@ async def connect(self) -> None:
...
             )
-        if self.server_info.schema_version > MIN_SCHEMA_VERSION:
-            # server version is higher than expected, log only
+        if info.schema_version > MAX_SCHEMA_VERSION:
+            # server version is too high, raise exception
+            await self._ws_client.close()
+            raise InvalidServerVersion(
+                f"Matter Server schema version is incompatible: {info.schema_version}"
+            )
+        if self.server_info.schema_version != SCHEMA_VERSION:
```
balloob, line 93: Should we even allow this? It sounds like it will just break as data models won't align ?

matter_server/common/const.py
```diff
@@ -0,0 +1,12 @@
+"""Constants that are shared between server and client."""
+
+# schema version is used to determine compatibility between server and client
+SCHEMA_VERSION = 1
+
+# we allow the schema version to be one version behind
+MIN_SCHEMA_VERSION = SCHEMA_VERSION - 1
```
balloob, line 7: I don't think that we have to make this dynamic, we could maybe allow 3 versions if they are minor extensions?

matter_server/common/const.py
```diff
@@ -0,0 +1,12 @@
...
+
+# we allow the schema version to be 2 versions ahead
```
balloob, line 9: Same.

matter_server/common/const.py
balloob: So basically, we just support as much as we can. Sometimes method signatures change a bit over time but it's not that much.

matter_server/client/connection.py
```diff
@@ -75,31 +75,23 @@ async def connect(self) -> None:
...
-                f"Matter Server schema version is incompatible: {info.schema_version} "
-                f"a version is required that supports at least api schema {MIN_SCHEMA_VERSION} "
-                " - update the Matter Server to a more recent version."
+                f"Matter schema version is incompatible: {info.schema_version}, "
+                f"the server requires at least {info.min_supported_schema_version} "
+                " - update the Matter client to a more recent version or downgrade the server."
             )
-        if self.server_info.schema_version > MIN_SCHEMA_VERSION:
-            # server version is higher than expected, log only
+        if abs(self.server_info.schema_version - SCHEMA_VERSION) > 1:
```
balloob, line 87: I think we can just drop this check? We’re either supported or not?

balloob (approved):

matter_server/client/connection.py
marcelveldt: yeah, I was trying to be too save. Removed it as it does not add any value

PR #213, opened 2023-02-21
Simplify datamodel

marcelveldt:
- Only store bare minimum of a node in the server (just the attributes raw value)
- Construct MatterNode in the client
- Construct DeviceType(instance) in the client
- Move Interview Logic 
- Bump schema version

BREAKING CHANGE

Marked as draft because this depends on updated wheels


matter_server/client/models/node.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def has_cluster(
+        self, cluster: type[_CLUSTER_T] | int, endpoint: int | None = None
+    ) -> bool:
+        """Check if node has a specific cluster on any of the endpoints."""
+        return any(
+            x
+            for x in self.endpoints.values()
+            if x.has_cluster(cluster)
+            and (x.endpoint_id == endpoint or endpoint is None)
```
MartinHjelmare, line 214: Start with the None check. Then we can save one attribute access if the check short-circuits.

matter_server/server/device_controller.py
```diff
@@ -71,18 +69,21 @@ async def initialize(self) -> None:
...
+            except (TypeError, KeyError, AttributeError) as err:
+                # only accept this to happen if the schema changed
+                if node_dict.get("interview_version") == SCHEMA_VERSION:
+                    raise err
+                # set node to None will be automatically re-interview the node
+                node = None
+            self._nodes[node_id] = node
+        # setup subscriptions and (re)interviews as task in the background
+        # as we do not want it to block our startup
+        asyncio.create_task(self._check_subscriptions_and_interviews())
```
MartinHjelmare, line 86: We need to store a reference to tasks to make sure they are not garbage collected.

marcelveldt: Also for short running tasks ? As this one reschedules itself.

balloob: yep

matter_server/server/device_controller.py
```diff
@@ -417,47 +443,67 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
+                except NodeInterviewFailed as err:
+                    LOGGER.warning(
+                        "Unable to interview Node %s, we will retry later in the background.",
+                        node_id,
+                        exc_info=err,
+                    )
+                    continue
+
+            # setup subscriptions for the node
+            if node_id not in self._subscriptions:
```
balloob, line 466: Turn this around.

```
if node_id in self._subscriptions:
    continue
```

matter_server/server/device_controller.py
```diff
@@ -417,47 +443,67 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
+
         # reschedule self to run every hour
-        self.server.loop.call_later(3600, self._schedule_interviews)
+        def _schedule():
+            asyncio.create_task(self._check_subscriptions_and_interviews())
 
-    def _schedule_interviews(self) -> None:
-        """Schedule interviews."""
-        asyncio.create_task(self._check_interviews())
+        # NOTE: do not do this in one call otherwise asyncio will start
```
balloob, line 485: Is this note really necessary? That's how asyncio works 😄

matter_server/server/device_controller.py
```diff
@@ -337,8 +350,8 @@ def attribute_updated_callback(
             new_value = transaction.GetAttribute(path)
             LOGGER.debug("attribute updated -- %s - new value: %s", path, new_value)
```
balloob, line 351: should this log include node ID ?

matter_server/server/device_controller.py
marcelveldt: hahaha, well, I've seen the mistake happen a lot.

matter_server/server/device_controller.py
marcelveldt: ah good to know, I'll adjust. For my knowledge: was that changed in python3.11 ?

matter_server/server/device_controller.py
marcelveldt: in fact that whole log line can be removed, I used it for debugging as I couldn't use a breakpoint here due to that callback called by the SDK

marcelveldt: Adjusted the logging a bit so the node events get logged to their own log child. Figured that logging might still come in handy

marcelveldt:
Next step is build a new SDSK clusters wheel which includes the ALL_CLUSTERS and ALL_ATTRIBUTES lookup dicts.

marcelveldt:
wheels are ready and I've addressed the final bugs/feedback.

MartinHjelmare:
There are some lint failures from mypy.

balloob (approved):

PR #217, opened 2023-02-22
Prepare bridge support

marcelveldt:
Flattened the datamodel a bit more as we've discovered that the datamodel was not flexible enough to cover for bridges.
Better to handle extraction of nodes, clusters and attributes into devices and entities from HA itself with a discovery schema comparable to e.g. Zwave.

- Removed the devicetype and rootnode instance layers.
- Added some helper attributes to discover if you're dealing with a bridge/bridged device


matter_server/client/models/node.py
```diff
@@ -69,9 +84,12 @@ def get_cluster(self, cluster: type[_CLUSTER_T] | int) -> _CLUSTER_T | None:
 
         Return None if the Cluster is not present on the node.
         """
-        if isinstance(cluster, type):
-            return self.clusters[cluster.id]  # type: ignore[no-any-return]
-        return self.clusters[cluster]  # type: ignore[no-any-return]
+        try:
+            if isinstance(cluster, type):
+                return self.clusters[cluster.id]  # type: ignore[no-any-return]
+            return self.clusters[cluster]  # type: ignore[no-any-return]
```
balloob, line 90: this shouldn't happen, as it's always a type ?

matter_server/client/models/node.py
```diff
@@ -69,9 +84,12 @@ def get_cluster(self, cluster: type[_CLUSTER_T] | int) -> _CLUSTER_T | None:
...
+        except KeyError:
```
balloob, line 91: You can use `self.clusters.get(cluster.id)` instead of catching KeyError

matter_server/client/models/node.py
marcelveldt: mypy was complaining to me but could be old mess

matter_server/client/models/node.py
marcelveldt: I'll just do that instead yeah

matter_server/client/models/node.py
```diff
@@ -155,91 +173,64 @@ def set_attribute_value(self, attribute_path: str, attribute_value: Any) -> None
...
         setattr(cluster_instance, attribute_name, attribute_value)
 
+    def update(self, attributes_data: dict[str, Any]) -> None:
+        """Update MatterEndpoint from (endpoint-specific) raw Attributes data."""
+        # unwrap cluster and clusterattributes from raw node data attributes
+        for attribute_path, attribute_value in attributes_data.items():
+            self.set_attribute_value(attribute_path, attribute_value)
+        # extract device types from Descriptor Cluster
+        if cluster := self.get_cluster(Clusters.Descriptor):
+            for dev_info in cluster.deviceTypeList:  # type: ignore[unreachable]
```
balloob, line 183: Why would we do this every time?

matter_server/client/models/node.py
```diff
@@ -155,91 +173,64 @@ def set_attribute_value(self, attribute_path: str, attribute_value: Any) -> None
...
+                device_type = DEVICE_TYPES.get(dev_info.type)
+                if device_type is None:
+                    LOGGER.debug("Found unknown device type %s", dev_info)
+                    continue
+                self.device_types.add(device_type)
```
balloob, line 188: Should we also be able to remove device types ?

matter_server/client/models/node.py
```diff
@@ -155,91 +173,64 @@ def set_attribute_value(self, attribute_path: str, attribute_value: Any) -> None
...
+        Return device info for this Node.
+
+        Returns BasicInformation from the Node itself (endpoint 0).
+        """
+        return self.get_cluster(0, Clusters.BasicInformation)
+
+    @property
+    def is_bridge_device(self) -> bool:
+        """Return if this Node is a Bridge/Aggregator device."""
+        return any(Aggregator in x.device_types for x in self.endpoints.values())
```
balloob, line 229: Shouldn't that always be endpoint 0 ?

matter_server/client/models/node.py
```diff
@@ -270,18 +261,27 @@ def get_cluster(
...
-            ),
-        )
+    def update(self, node_data: MatterNodeData) -> None:
+        """Update MatterNode from MatterNodeData."""
+        self.node_data = node_data
+        # collect per endpoint data
+        endpoint_data: dict[int, dict[str, Any]] = {}
+        for attribute_path, attribute_data in node_data.attributes.items():
+            endpoint_id = int(attribute_path.split("/")[0])
+            endpoint_data.setdefault(endpoint_id, {})
```
balloob, line 268: maybe better to check if key exists instead of using `setdefault`, otherwise you're going to create and throw away a lot of empty dicts.

matter_server/client/models/node.py
marcelveldt: adjusted

balloob (approved):

matter_server/client/models/node.py
marcelveldt: no only once

marcelveldt: that update is only called if the entire node updates (e.g. after interview)

matter_server/client/models/node.py
```diff
@@ -155,91 +173,64 @@ def set_attribute_value(self, attribute_path: str, attribute_value: Any) -> None
         )
         setattr(cluster_instance, attribute_name, attribute_value)
 
+    def update(self, attributes_data: dict[str, Any]) -> None:
+        """Update MatterEndpoint from (endpoint-specific) raw Attributes data."""
+        # unwrap cluster and clusterattributes from raw node data attributes
+        for attribute_path, attribute_value in attributes_data.items():
+            self.set_attribute_value(attribute_path, attribute_value)
+        # extract device types from Descriptor Cluster
+        if cluster := self.get_cluster(Clusters.Descriptor):
```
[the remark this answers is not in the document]
balloob: casting is if type is unknown. If we need to give mypy a hint, use `assert cluster is not None`

matter_server/client/models/node.py
marcelveldt: I saw it at endpoint 1 on the Hue bridge

marcelveldt:
This one is ready, tested with HA (PR following up next)

matter_server/client/models/node.py
```diff
@@ -270,18 +272,56 @@ def get_cluster(
...
+        for endpoint_id, attributes_data in endpoint_data.items():
+            if endpoint_id in self.endpoints:
+                self.endpoints[endpoint_id].update(attributes_data)
+            else:
+                self.endpoints[endpoint_id] = MatterEndpoint(
+                    endpoint_id=endpoint_id, attributes_data=attributes_data, node=self
+                )
+        # lookup if this is a bridge device
+        self._is_bridge_device = any(
+            Aggregator in x.device_types for x in self.endpoints.values()
```
MartinHjelmare, line 304: Rename `x` to `endpoint` if that's what it is.

matter_server/client/models/node.py
```diff
@@ -270,18 +272,56 @@ def get_cluster(
...
         )
+        # composed devices reference to other endpoints through the partsList attribute
+        # create a mapping table to quickly map this
+        for endpoint in self.endpoints.values():
```
MartinHjelmare, line 308: Consider setting `_is_bridge_device` in this iteration instead of above.

PR #222, opened 2023-02-28
Fix open commissioning window

ArturoGuerra:
The open commissioning window function should now return a pairing code and a QR code, we just need to either wait for the next release of matter or patch it.

https://github.com/project-chip/connectedhomeip/blob/92d86d5afaa47446ebe6377d7350534da8ec3d84/src/controller/python/chip/ChipDeviceCtrl.py#L763-L770


agners:
Hm, this would need backporting of https://github.com/project-chip/connectedhomeip/pull/24560. We'd have to check how hard it is to backport that.

ArturoGuerra:
Yeah that does look like it would be quite a lot of work, maybe it's just worth waiting for the next matter release which if I'm not mistaken should be this month

marcelveldt:
Let's wait for the new Matter release which solved a few of our missing dependencies

MartinHjelmare:
Are all the required dependencies available now so this PR can go ahead?

agners (approved):
FWIW, this is related to this change: https://github.com/project-chip/connectedhomeip/pull/24560/files#diff-6a5e5e3a7f878219ea9863c0774cb5927828a4442c983aeadc58e99d661ebd20R770, which is part of CHIP wheels 1.1.0.1 we are using.

LGTM.

PR #226, opened 2023-03-04
Update README.md

codyc1515:
Add instructions on git cloning

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,8 @@ Please note that development is only possible on Linux and MacOS, no Windows sup
 ### Build
 
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -198,6 +198,8 @@ Please note that development is only possible on Linux and MacOS, no Windows sup
 ### Build
 
 ```sh
+git clone https://github.com/home-assistant-libs/python-matter-server.git
```
MartinHjelmare, line 201: This is only needed the first time you start developing. After that you need another command to keep the local repository synced with the remote repository.

I don't think we should add these commands as they are common for all development for git repositories.

PR #238, opened 2023-03-20
A few small fixes and minor improvements

marcelveldt:
- extend the heartbeat interval from 10 seconds to 120 seconds to reduce network traffic
- disable the fabric filter on the read/subscription request, this will allow managing fabrics*

* this will also bring the possibility to actually remove a device from our fabric


scripts/beautify_diagnostics.py
```diff
@@ -1,11 +1,13 @@
 """Script to beautify diagnostics output."""
 import json
+from os.path import abspath, dirname
 import sys
 
 from chip.clusters.ClusterObjects import ALL_ATTRIBUTES, ALL_CLUSTERS
 import yaml
 
-from matter_server.client.models.device_types import ALL_TYPES
+sys.path.insert(1, dirname(dirname(abspath(__file__))))
```
MartinHjelmare, line 9: Why is this needed?

marcelveldt: so you can run the script without the lib installed, just from your project root

marcelveldt: Meh, I don't want to on my dev machine

MartinHjelmare: Install with:

```
pip install -e .[server]
```

That will symlink the folder.

marcelveldt: Unfortunately that does not work on MacOS. It actually works when sqitching from zsh shell to bash but then the next issue is the wheels requirement which can not be met so for now we'll have to leave the workaround in place

MartinHjelmare (approved):

matter_server/client/models/node.py
```diff
@@ -279,7 +279,7 @@ def get_compose_parent(self, endpoint_id: int) -> MatterEndpoint | None:
         return None
 
     def get_compose_child_ids(self, endpoint_id: int) -> tuple[int, ...] | None:
-        """Return endpoint ID's of any childs if the endpoint represents a Composed device."""
+        """Return endpoint ID's of any child's if the endpoint represents a Composed device."""
```
MartinHjelmare, line 282:
```suggestion
        """Return endpoint IDs of any child if the endpoint represents a Composed device."""
```

scripts/beautify_diagnostics.py
```diff
@@ -5,7 +5,7 @@
 from chip.clusters.ClusterObjects import ALL_ATTRIBUTES, ALL_CLUSTERS
 import yaml
 
-from matter_server.client.models.device_types import ALL_TYPES
+from matter_server.client.models.device_types import ALL_TYPES  # noqa: E402
```
MartinHjelmare, line 8: Is the noqa needed?

marcelveldt: nope, leftover

PR #240, opened 2023-03-22
Automatically retrieve PAA certificates for device attestation

marcelveldt:
For commissioning a device its PAA certificate is required to do device attestation checks.
With the certificate missing, device commissioning will fail. Now that Matter devices are hitting the market every week, we'd like to keep the certificates up to date.

This change fetches all certificates at startup.
When starting a commissioning, it will do  quick delta sync of certificates to ensure we have the latest set.

The full sync at startup happens before initializing the device controller.
For the delta sync at commissioning-stage I'm not so sure that any new certificates get picked up by the device controller. By some basic testing, it seemed to me that the device controller needs a restart/reload in order to pick up new certificates. This is a bit tricky due all the C bindings so I did not implement that for now.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -7,12 +7,15 @@ from collections import deque
 from datetime import datetime
 from functools import partial
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Deque, Type, TypeVar, cast
+import pathlib
+from typing import TYPE_CHECKING, Any, Callable, Deque, Final, Type, TypeVar, cast
 
 from chip.ChipDeviceCtrl import CommissionableNode
 from chip.clusters import Attribute
 from chip.clusters.ClusterObjects import ALL_CLUSTERS, Cluster
 from chip.exceptions import ChipStackError
+
+from matter_server.server.helpers.paa_certificates import fetch_certificates
 
 from ..common.const import SCHEMA_VERSION
 from ..common.errors import (
@@ -35,6 +38,18 @@ DATA_KEY_LAST_NODE_ID = "last_node_id"
 
 LOGGER = logging.getLogger(__name__)
 
+# the paa-root-certs path is hardcoded in the sdk at this time
+# and always uses the development subfolder
+# regardless of anything you pass into instantiating the controller
+# revisit this once matter 1.1 is released
+PAA_ROOT_CERTS_DIR: Final[pathlib.Path] = (
+    pathlib.Path(__file__)
+    .parent.resolve()
+    .parent.resolve()
+    .parent.resolve()
+    .joinpath("credentials/development/paa-root-certs")
+)
+
 
 class MatterDeviceController:
     """Class that manages the Matter devices."""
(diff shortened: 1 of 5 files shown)


matter_server/server/device_controller.py
```diff
@@ -46,7 +61,8 @@ def __init__(
         """Initialize the device controller."""
         self.server = server
         # Instantiate the underlying ChipDeviceController instance on the Fabric
-        self.chip_controller = server.stack.fabric_admin.NewController()
+        assert PAA_ROOT_CERTS_DIR.is_dir()
```
MartinHjelmare, line 66: Maybe raise `RuntimeError` if it's missing instead? Asserts can be turned off.

MartinHjelmare: This will do I/O, btw, so we should schedule it on the executor.

matter_server/server/device_controller.py
```diff
@@ -56,13 +72,14 @@ def __init__(
...
-    @property
-    def fabric_id(self) -> int:
-        """Return Fabric ID."""
-        return cast(int, self.chip_controller.fabricId)
-
     async def initialize(self) -> None:
         """Async initialize of controller."""
+        # (re)fetch all PAA certificates once at startup
+        # NOTE: this must be done before initializing the controller
+        asyncio.create_task(fetch_certificates(PAA_ROOT_CERTS_DIR))
```
MartinHjelmare, line 81: Shouldn't we await fetching the certificates?

matter_server/server/helpers/paa_certificates.py
```diff
@@ -0,0 +1,103 @@
...
+    if fetch_production_certificates:
+        base_urls.add(PRODUCTION_URL)
+
+    async with ClientSession() as http_session:
+        for url_base in base_urls:
+            # fetch the paa certificates list
+            async with http_session.get(
+                f"{url_base}/dcl/pki/root-certificates"
+            ) as response:
+                assert response.status == 200
```
MartinHjelmare, line 76: I'd use raise for errors and raise a specific error out of that, that the caller can handle.

matter_server/server/device_controller.py
marcelveldt: This is in the init where we do a lot more IO before we start all async stuff

matter_server/server/device_controller.py
marcelveldt: oops, yes, copy paste error

matter_server/server/helpers/paa_certificates.py
marcelveldt: I have no idea what can go wrong here. I'm tempted to just catch all aiohttp errors here, log it and skip

matter_server/server/device_controller.py
MartinHjelmare: Right, I see it's before we start the server. 👍

matter_server/server/helpers/paa_certificates.py
```diff
@@ -0,0 +1,107 @@
...
+                    await write_paa_root_cert(
+                        certificate,
+                        subject,
+                        paa_trust_store_path,
+                    )
+                    LAST_CERT_IDS.add(paa["subjectKeyId"])
+                    fetch_count += 1
+    except ClientError as err:
+        LOGGER.warning(
+            "Fetching latest certificates failed: error %s", str(err), exc_info=err
```
MartinHjelmare, line 103:
```suggestion
            "Fetching latest certificates failed: error %s", err, exc_info=err
```

MartinHjelmare (approved):

marcelveldt:
> It doesn't look like this passed the CI checks successfully.

It was green, maybe GH slow ?

PR #247, opened 2023-03-29
Fix lint and test issues on main

MartinHjelmare:
- Address issues after https://github.com/home-assistant-libs/python-matter-server/pull/240.
- Move the certificates path constant to server constants. Use this constant directly in the helper script to fetch and generate certificates.
- Instantiate the chip controller when initializing the device controller instead of when instantiating the device controller.


marcelveldt (approved):
Thanks!

PR #249, opened 2023-04-02
Fabric management

ArturoGuerra:
This will add client and server functions to help with the management of matter fabrics and the mapping of CSA vendor ids with vendor names.


ArturoGuerra:
For the time being the fetching of vendor info from the CSA is done on the server side and stored in memory as it's not that much information. So we could either change it so it's stored persistently and updated at startup or move it to the client side.

matter_server/client/client.py
```diff
@@ -156,6 +157,49 @@ async def open_commissioning_window(
             ),
         )
 
+    async def get_matter_fabrics(
+        self, node_id: int
+    ) -> list[tuple[int, int, int, str | None, str | None]]:
+        """
+        Get Matter fabrics from a device.
+
+        Returns a list of tuples containing:
```
marcelveldt, line 166: maybe create a simple model (dataclass) for this ?

ArturoGuerra: Yeah it initially didn't return that much data ill fix that

matter_server/client/client.py
```diff
@@ -156,6 +157,47 @@ async def open_commissioning_window(
             ),
         )
 
+    async def get_matter_fabrics(self, node_id: int) -> list[MatterFabricData]:
+        """
+        Get Matter fabrics from a device.
+
+        Returns a list of tuples containing:
```
marcelveldt, line 164: docstring needs to be adjusted to the new model

marcelveldt (approved):
just that minor docstring comment, other than that: very nice !

matter_server/client/client.py
```diff
@@ -156,6 +157,46 @@ async def open_commissioning_window(
...
 
+    async def get_matter_fabrics(self, node_id: int) -> list[MatterFabricData]:
+        """
+        Get Matter fabrics from a device.
+
+        Returns a list of MatterFabricData objects.
+        """
+
+        node = await self.get_node(node_id)
+        fabric: list[
```
marcelveldt, line 168: should't this be fabrics ?

matter_server/server/vendor_info.py
```diff
@@ -0,0 +1,97 @@
...
+        LOGGER.info("Saving vendor info to storage.")
+        self._server.storage.set(
+            DATA_KEY_VENDOR_INFO,
+            {
+                vendor_id: dataclass_to_dict(vendor_info)
+                for vendor_id, vendor_info in self._data.items()
+            },
+        )
+
+    @api_command(APICommand.GET_VENDOR_NAMES)
```
MartinHjelmare, line 82: I think a new command requires a bump of schema version. The server needs to support this command if a client wants to send the command to the server.

marcelveldt: sharp as always Martin. I'll bump the schema

PR #258, opened 2023-04-08
Add missing test VID certificates

ArturoGuerra:
This will add the required test VID certificates from the matter git but in the future, we should add a way to change this behavior from the UI, like that the end user can choose which certificates to allow.

Fixes https://github.com/home-assistant-libs/python-matter-server/issues/255


marcelveldt (approved):
nice, thanks!

matter_server/server/helpers/paa_certificates.py
```diff
@@ -21,7 +21,11 @@
 LOGGER = logging.getLogger(__name__)
 PRODUCTION_URL = "https://on.dcl.csa-iot.org"
 TEST_URL = "https://on.test-net.dcl.csa-iot.org"
-
+GIT_URL = "https://github.com/project-chip/connectedhomeip/raw/master/credentials/development/paa-root-certs"  # pylint: disable=line-too-long
```
MartinHjelmare, line 24: Please split long strings on multiple lines using parenthesis.

PR #264, opened 2023-04-12
Changed node removal behavior so fabric removal is optional

ArturoGuerra:
This change will make sure we always remove a node, even if it's offline or more importantly has been removed from our fabric by another controller or has been reset by the user.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -376,13 +376,16 @@ class MatterDeviceController:
         )
         fabric_index = node.attributes[attribute_path]
 
-        await self.chip_controller.SendCommand(
-            nodeid=node_id,
-            endpoint=0,
-            payload=Clusters.OperationalCredentials.Commands.RemoveFabric(
-                fabricIndex=fabric_index,
-            ),
-        )
+        try:
+            await self.chip_controller.SendCommand(
+                nodeid=node_id,
+                endpoint=0,
+                payload=Clusters.OperationalCredentials.Commands.RemoveFabric(
+                    fabricIndex=fabric_index,
+                ),
+            )
+        except ChipStackError as err:
+            LOGGER.error("Failed to remove node %s from fabric: %s", node_id, err)
 
         self.server.signal_event(EventType.NODE_DELETED, node_id)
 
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):

matter_server/server/device_controller.py
```diff
@@ -376,13 +376,16 @@ async def remove_node(self, node_id: int) -> None:
...
-        )
+        try:
+            await self.chip_controller.SendCommand(
+                nodeid=node_id,
+                endpoint=0,
+                payload=Clusters.OperationalCredentials.Commands.RemoveFabric(
+                    fabricIndex=fabric_index,
+                ),
+            )
+        except ChipStackError as err:
```
MartinHjelmare, line 387: I don't understand the PR description. The node is popped from the dict already above the call.

We're only ensuring the event is signaled. But we no longer raise so the caller won't be notified of the failure.

ArturoGuerra: Yeah so basically the current behavior is you just can't remove the device on the HA side, it stays there and it doesn't get removed until the matter server is restarted. The current idea with this pull request is to signal HA to remove the device and ignore the fact we couldn't remove the fabric which is not necessarily a bad thing but I guess we could go back to the previous behavior and move the signaling so it happens before we call the sdk.

MartinHjelmare: Yes, that sounds good. Always send the signal, and raise on failure.

balloob: Should we go one step further: keep track that it failed but should be removed, and if the device ever pops up again, we remove it then

ArturoGuerra: We could do that but I really don't think that would be necessary as nobody else is doing it, from what I've seen apple and google both try one time to remove their fabric from the device and if they fail that's it, and it could also cause issues if we have a lot of devices that have been removed and we just keep on retrying so we would have to also implement an expiration system for that.

balloob: I wouldn't keep retrying. But when we get notified of data from a node we haven't seen this run before, we could check if we know it or if we tried deleting it in the past.

marcelveldt: we will not/never receive any node events if we do not resolve/subscribe it so that is a bit a case of the chicken and the egg.

marcelveldt (approved):
I think its fine like this, keep it simple for now as the most common usecase is removing a node after reset anyways.
Thanks!

PR #266, opened 2023-04-13
Process nodes asynchronously

ArturoGuerra:
This will process node interviews and subscriptions asynchronously and block from running the task again until all queued nodes have been processed.

fixes: https://github.com/home-assistant-libs/python-matter-server/issues/265

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -7,7 +7,7 @@ from collections import deque
 from datetime import datetime
 from functools import partial
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Deque, Type, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Callable, Coroutine, Deque, Type, TypeVar, cast
 
 from chip.ChipDeviceCtrl import CommissionableNode
 from chip.clusters import Attribute, Objects as Clusters
@@ -42,6 +42,7 @@ DATA_KEY_NODES = "nodes"
 DATA_KEY_LAST_NODE_ID = "last_node_id"
 
 LOGGER = logging.getLogger(__name__)
+INTERVIEW_TASK_LIMIT = 5
 
 
 class MatterDeviceController:
@@ -527,8 +528,8 @@ class MatterDeviceController:
 
     async def _check_subscriptions_and_interviews(self) -> None:
         """Run subscriptions (and interviews) for known nodes."""
-        interview_tasks: list[asyncio.Task] = []
-        subscription_tasks: list[asyncio.Task] = []
+        tasks: list[Coroutine[Any, Any, None]] = []
+        task_limit: asyncio.Semaphore = asyncio.Semaphore(INTERVIEW_TASK_LIMIT)
         for node_id, node in self._nodes.items():
             # (re)interview node (only) if needed
             if (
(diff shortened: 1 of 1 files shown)


ArturoGuerra:
Also is there any reason we process nodes hourly? wouldn't that make it so anything that was offline during startup wait a whole hour before it can be used?

matter_server/server/device_controller.py
```diff
@@ -527,38 +527,27 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
-                    await self.interview_node(node_id)
-                except NodeInterviewFailed as err:
-                    LOGGER.warning(
-                        "Unable to interview Node %s, we will retry later in the background.",
-                        node_id,
-                        exc_info=err,
-                    )
-                    continue
+                interview_tasks.append(
+                    asyncio.create_task(self.interview_node(node_id))
```
MartinHjelmare, line 540: We should probably still catch `NodeInterviewFailed` so we need a wrapping function that can do that and that we can schedule as a task.

matter_server/server/device_controller.py
```diff
@@ -527,38 +527,27 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
-                # If the node is unreachable on the network now,
-                # it will throw a NodeNotResolving exception, catch this,
-                # log this and just try to resolve this node in the next run.
-                LOGGER.warning(
-                    "Unable to contact Node %s,"
-                    " we will retry later in the background.",
-                    node_id,
-                    exc_info=err,
-                )
+            subscription_tasks.append(asyncio.create_task(self.subscribe_node(node_id)))
```
MartinHjelmare, line 547: Same for `NodeNotResolving`.

marcelveldt:
I actually did that on purpose because the ResolveNode was a bit fragile when called simultane so interview them one by one at startup (if needed) and run a task every hour to check if there are any nodes that need a (re)interview.

So... this needs real proper testing. It would be great if the SDK can now handle multiple resolve and interviews commands at the same time  but in my experience that is not the case. 

Also for thread devices this causes A LOT of traffic so I am not a fan of doing this async unless we have some lock on the interview/read call

marcelveldt:
As discussed on discord: 
- semaphore for max 5 interviews/read requests at the same time
- assume that we now can call resolvenode multiple times simultanuous

matter_server/server/device_controller.py
```diff
@@ -527,38 +529,75 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
+            tasks.append(_subscribe_node(node_id))
+
+        async def _run_task(task: Coroutine[Any, Any, None]) -> None:
+            """Run coroutine and release semaphore."""
+            async with task_limit:
+                await task
+
+        LOGGER.debug("Running %s tasks", len(tasks))
+        # wait for all tasks to finish
+        futures: list[asyncio.Future] = await asyncio.gather(
```
MartinHjelmare, line 587:
```suggestion
        results: list[Exception | None] = await asyncio.gather(
```

matter_server/server/device_controller.py
```diff
@@ -527,38 +529,75 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
+            *(_run_task(task) for task in tasks), return_exceptions=True
+        )
+        LOGGER.debug(
+            "Done running %s tasks in %s seconds",
+            len(futures),
```
MartinHjelmare, line 592:
```suggestion
            len(results),
```

matter_server/server/device_controller.py
```diff
@@ -527,38 +529,75 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
+            start_time - time.time(),
+        )
+        # check if any of the tasks failed
+        for future in futures:
```
MartinHjelmare, line 596:
```suggestion
        for result in results:
```

matter_server/server/device_controller.py
```diff
@@ -527,38 +529,75 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
+            if isinstance(future, Exception):
```
MartinHjelmare, line 597:
```suggestion
            if isinstance(result, Exception):
```

MartinHjelmare (changes requested):

MartinHjelmare (approved):
Marcel should approve too.

marcelveldt (approved):
yes! well done!

marcelveldt:
works like a charm!

PR #297, opened 2023-05-23
Handle edgecase where descriptor cluster is missing

marcelveldt:
Strictly speaking this may not happen, yet during testing with Switchbot devices I came across the edge case that one of the endpoints does not have a Descriptor cluster at all (completely missing).

Handle this in the code so we do not crash when this happens.

### DIFF
--- a/matter_server/client/models/node.py
+++ b/matter_server/client/models/node.py
@@ -197,14 +197,13 @@ class MatterEndpoint:
         for attribute_path, attribute_value in attributes_data.items():
             self.set_attribute_value(attribute_path, attribute_value)
         # extract device types from Descriptor Cluster
-        cluster = self.get_cluster(Clusters.Descriptor)
-        assert cluster is not None
-        for dev_info in cluster.deviceTypeList:  # type: ignore[unreachable]
-            device_type = DEVICE_TYPES.get(dev_info.deviceType)
-            if device_type is None:
-                LOGGER.debug("Found unknown device type %s", dev_info)
-                continue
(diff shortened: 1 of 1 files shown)


matter_server/client/models/node.py
```diff
@@ -197,14 +197,13 @@ def update(self, attributes_data: dict[str, Any]) -> None:
...
-        for dev_info in cluster.deviceTypeList:  # type: ignore[unreachable]
-            device_type = DEVICE_TYPES.get(dev_info.deviceType)
-            if device_type is None:
-                LOGGER.debug("Found unknown device type %s", dev_info)
-                continue
-            self.device_types.add(device_type)
+        if cluster := self.get_cluster(Clusters.Descriptor):
+            for dev_info in cluster.deviceTypeList:  # type: ignore[unreachable]
+                device_type = DEVICE_TYPES.get(dev_info.deviceType)
+                if device_type is None:
```
hidaris, line 203: If a device type is not specified, the device will be considered unverified. Should we provide support for this type of device?

marcelveldt: for now yes, just to not crash. I've already emailed Switchbot about the issue.

agners: I'd prefer if we at least write something to the logs. Seems steep going from assert to silently ignore.

marcelveldt: that assert was there for the type checker. I do not see any real benefit of logging here for such an edge case.

PR #325, opened 2023-06-22
Validate GitHub release on publish

agners:
Make sure the tag created by the GitHub release matches the Python project metadata.

This is in preparation for Docker image build, which will rely on the release tag to fetch the latest Matter Server from PyPI.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -2,7 +2,7 @@ name: Publish releases to PyPI
 
 on:
   release:
-    types: [published, prereleased]
+    types: [published]
 
 jobs:
   build-and-publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -2,21 +2,31 @@ name: Publish releases to PyPI
...
       - name: Install build
         run: >-
           pip install build
+      - name: Validate tag matches Python project version
+        shell: python
+        run: |-
+          import tomli
+          with open("pyproject.toml", "rb") as f:
+            pyproject = tomli.load(f)
+          assert pyproject["project"]["version"] == "${{ steps.vars.outputs.tag }}"
```
frenck, line 29: Alternatively (and IMHO better solution), is to not set it and update it on release automatically.

agners: Hm, so that would involve reading pyproject.toml here, setting the version and writing it back out?

agners: I've implemented that in a new PR, see #326.

PR #326, opened 2023-06-22
Use release tag as project version when creating a release

agners:
Use the tag created by the GitHub release and write it in the Python project metadata.

This is in preparation for Docker image build, which will rely on the release tag to fetch the latest Matter Server from PyPI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-matter-server"
-version = "3.5.1"
+version = "0.0.0"
 license     = {text = "Apache-2.0"}
 description = "Python Matter WebSocket Server"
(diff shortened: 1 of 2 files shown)


marcelveldt:
This has a side-effect that breaks the development environment due to the version check

marcelveldt (approved):
nice !

One comment is maybe put a comment in the projectfile that we dynamically generate the version from the github release ?

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-matter-server"
-version = "3.5.1"
+version = "0.0.0"
```
agners, line 7:
```suggestion
# The version is set by GH action on release
version = "0.0.0"
```

PR #339, opened 2023-06-28
Some small fixes for the client

marcelveldt:
Some small fixes/improvements based on feedback last couple of days:

- Require schema version 4 for the call to "subscribe_attribute"
- Rename "subscribe" to "subscribe_events" to make its purpose more clear [BREAKING]
- Raise NodeNotExists instead of KeyError in get_node method if the node does not exist


MartinHjelmare (approved):
PR description needs an update.

PR #341, opened 2023-06-28
Add feature to write an attribute to a node

### LINKED ISSUE #285: How to write attributes from clients
Hi
I'm developing fan control integration, but this cluster don't have any commands to set speed. I would like to set speed attribute but python-matter-server don't support any API to do it.
Similarly, I think we should have `APICommand.WRITE_ATTRIBUTE`
```
@api_command(APICommand.WRITE_ATTRIBUTE)
    async def write_attribute(self, node_id: int, endpoint: int, attribute: ClusterAttributeDescriptor, timedRequestTimeoutMs: int = None, interactionTimeoutMs: int = None):
        """Write a attribute on a target node."""
        LOGGER.info("Write Attributes node_id: %s", node_id)
        attributes = [(endpoint, attribute)]
        return await self.chip_controller.WriteAttribute(
            nodeid=node_id,
            attributes=attributes,
            timedRequestTimeoutMs=timedRequestTimeoutMs,
            interactionTimeoutMs=interactionTimeoutMs,
        )
```

marcelveldt:
Fixes #285

### DIFF
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -242,6 +242,7 @@ class MatterClient:
         """Write an attribute(value) on a target node."""
         return await self.send_command(
             APICommand.WRITE_ATTRIBUTE,
+            require_schema=4,
             node_id=node_id,
             attribute_path=attribute_path,
             value=value,
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -240,7 +240,7 @@ class MatterClient:
         timed_request_timeout_ms: int | None = None,
         interaction_timeout_ms: int | None = None,
     ) -> Any:
-        """Send a command to a Matter node/device."""
+        """Write an attribute(value) on a target node."""
         return await self.send_command(
             APICommand.WRITE_ATTRIBUTE,
             node_id=node_id,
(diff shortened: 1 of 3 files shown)


matter_server/client/client.py
```diff
@@ -231,6 +231,28 @@ async def send_device_command(
             interaction_timeout_ms=interaction_timeout_ms,
         )
 
+    async def write_attribute(
```
MartinHjelmare, line 234: What features need to write attributes? Why doesn't all features that does a change have a device command?

marcelveldt: commands that are not predefined in the clusters, manufacturer specific data.
For example setting the mode of a thermostat is not defined as cluster command but thermostats can accept it with writing an attribute

marcelveldt: see chapter 8.9.2.4 of the specs btw

matter_server/server/device_controller.py
```diff
@@ -383,6 +383,24 @@ async def send_device_command(
                 interactionTimeoutMs=interaction_timeout_ms,
             )
 
+    @api_command(APICommand.WRITE_ATTRIBUTE)
+    async def write_attribute(
+        self,
+        node_id: int,
+        attribute_path: str | list[str],
```
MartinHjelmare, line 390: Maybe only support a single path? Is it common to write the same value to multiple paths at the same time?

MartinHjelmare (approved):

PR #342, opened 2023-06-28
Add read attribute command/feature

marcelveldt:
Add a command to read a live attribute value from a node

Usecase is a one-off read of some value (Like the thread diagnostics) where a subscription is not needed/wanted.

### DIFF
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -231,6 +231,19 @@ class MatterClient:
             interaction_timeout_ms=interaction_timeout_ms,
         )
 
+    async def read_attribute(
+        self,
+        node_id: int,
+        attribute_path: str | list[str],
+    ) -> Any:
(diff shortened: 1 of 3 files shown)


matter_server/client/client.py
```diff
@@ -231,6 +233,19 @@ async def send_device_command(
             interaction_timeout_ms=interaction_timeout_ms,
         )
 
+    async def read_attribute(
+        self,
+        node_id: int,
+        attribute_path: str | list[str],
```
MartinHjelmare, line 239: Should it be a list or just a single path?

marcelveldt: the same typo again, adjusted it

MartinHjelmare (approved):

MartinHjelmare:
Just solve the merge conflicts.

PR #346, opened 2023-06-29
Automatic versioning with release drafter

marcelveldt:
Automatic version bumps to major, minor or patch based on the labels used within PR's.

@ludeeus @MartinHjelmare

### DIFF
--- a/.github/workflows/pr-labels.yaml
+++ b/.github/workflows/pr-labels.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 🏷 Verify PR has a valid label
-        uses: ludeeus/action-require-labels@main
+        uses: ludeeus/action-require-labels@1.0.0
         with:
          labels: >-
             breaking-change, bugfix, refactor, new-feature, maintenance, ci, dependencies
--- a/.github/workflows/pr-labels.yaml
+++ b/.github/workflows/pr-labels.yaml
@@ -14,11 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 🏷 Verify PR has a valid label
-        uses: jesusvasquez333/verify-pr-label-action@v1.4.0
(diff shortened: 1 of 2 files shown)


marcelveldt:
Also added the workflow for the PR labels to be mandatory

.github/workflows/pr-labels.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+# yamllint disable-line rule:truthy
+on:
+  pull_request_target:
+    types:
+      - opened
+      - labeled
+      - unlabeled
+      - synchronize
+  workflow_call:
```
ludeeus, line 12:
```suggestion
    types:
      - labeled
      - unlabeled
```
These are not needed here

MartinHjelmare (approved):
Side note: It would perhaps be good to be able to run the release-drafter workflow manually via `workflow_dispatch` event? That way we can re-label PRs and re-run release-drafter.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

ludeeus:
That is one option, you can also just rerun the last run (its the same place as you have to trigger it anyway)

marcelveldt:
So is this PR fine as-is ? We don't adjust the workflow for the labels (and accept the bit of overhead it brings) ?

.github/workflows/pr-labels.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: PR Labels
+
+# yamllint disable-line rule:truthy
+on:
+  pull_request_target:
+    types:
+      - labeled
+      - unlabeled
```
ludeeus, line 9:
```suggestion
on:
  pull_request:
    types:
      - synchronize
      - labeled
      - unlabeled
    branches:
      - main
```

ludeeus (approved):

MartinHjelmare (approved):

PR #355, opened 2023-07-10
Use stable tag in example compose file

agners:
Users are supposed to use the latest stable version of the container which is available under the tag "stable". The tag "latest" currently doesn't exist.


frenck (approved):
👍 

../Frenck

PR #357, opened 2023-07-11
Improve node resolving for nodes offline at startup

marcelveldt:
- Prevent concurrent resolve actions with a lock
- Simplify retry logic
- Retry more often

When one or more nodes are offline/unavailable (e.g. powered off, out of range, out of battery) at server startup, no automatic subscription is set-up for that device. Because the SDK does not expose a way to add a callback when a known node is discovered on mDNS, we do a regular poll to see if the device is back alive. This is now fixed and the poll interval is a little bit increased. It will start with a 30 second interval, slowly increasing that with 10 seconds to a maximum of 10 minutes, meaning that if one of those devices comes back online, it will take max 10 minutes for the server to pick the device back up again.

Once a device subscription has been set-up, a resubscription is done within a few seconds by the SDK.
It will however take 3 minutes before we detect that a device is offline which seems like a fair tradeoff between traffic vs comfort.

If we want to improve this even further, we may consider implementing zeroconf discovery within the server to actively listen for mDNS broadcasts for nodes we're watching but as this is a bit of an edge case (the device offline just while restarting the server) it may not be the highest priority.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -84,6 +84,7 @@ class MatterDeviceController:
         self._interview_limit: asyncio.Semaphore = asyncio.Semaphore(
             INTERVIEW_TASK_LIMIT
         )
+        self._resolve_lock: asyncio.Lock = asyncio.Lock()
         self._node_lock: dict[int, asyncio.Lock] = {}
 
     async def initialize(self) -> None:
@@ -744,7 +745,7 @@ class MatterDeviceController:
         )
 
     async def _check_interview_and_subscription(
-        self, node_id: int, reschedule_interval: int = 300
+        self, node_id: int, reschedule_interval: int = 30
     ) -> None:
         """Handle interview (if needed) and subscription for known node."""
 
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -837,37 +838,23 @@ def _parse_attributes_from_read_result(
...
         except (ChipStackError, TimeoutError) as err:
-            if not retries:
+            if retries <= 1:
                 # when we're out of retries, raise NodeNotResolving
                 raise NodeNotResolving(f"Unable to resolve Node {node_id}") from err
-            async with node_lock:
-                await self._resolve_node(
-                    node_id=node_id, retries=retries - 1, allow_pase=retries - 1 == 0
-                )
+            await self._resolve_node(node_id=node_id, retries=retries - 1)
```
agners, line 857: I think I'd prefer a while loop instead of recursion but since we had it already I am fine with it.

agners (approved):

matter_server/server/device_controller.py
```diff
@@ -837,37 +838,23 @@ def _parse_attributes_from_read_result(
...
         node_lock = self._get_node_lock(node_id)
         if self.chip_controller is None:
             raise RuntimeError("Device Controller not initialized.")
         try:
-            if allow_pase:
-                # last attempt allows PASE connection (last resort)
-                LOGGER.debug(
-                    "Attempting to resolve node %s (with PASE connection)", node_id
+            async with node_lock, self._resolve_lock:
+                LOGGER.info("Attempting to resolve node %s...", node_id)
```
MartinHjelmare, line 848: I'd log at debug level.

PR #369, opened 2023-07-25
A few small optimizations

marcelveldt:
Some small speedups and code cleanup

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -588,8 +588,9 @@ class MatterDeviceController:
                 # In order to prevent network congestion due to wildcard subscriptions on all nodes,
                 # we keep a list of attributes we are explicitly interested in.
                 attributes=attr_subscriptions,
-                # simply subscribe to all (urgent and non urgent) device events
-                events=[("*", 1), ("*", 0)],
+                # simply subscribe to urgent device events only (e.g. button press etc.)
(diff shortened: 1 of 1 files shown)


frenck (approved):

matter_server/server/device_controller.py
```diff
@@ -409,7 +403,7 @@ async def read_attribute(
                 )
             )
             read_atributes = self._parse_attributes_from_read_result(result.attributes)
-            return read_atributes[attribute_path]
+            return read_atributes.get(attribute_path, None)
```
MartinHjelmare, line 406: `None` is the default value returned when the key is missing in the dict.

matter_server/server/device_controller.py
```diff
@@ -856,10 +851,16 @@ def _parse_attributes_from_read_result(
...
+        if (node := self._nodes.get(node_id)) and node.available:
+            # no need to resolve, the node is already available/connected
+            return
+
         node_lock = self._get_node_lock(node_id)
         if self.chip_controller is None:
             raise RuntimeError("Device Controller not initialized.")
         try:
+            # the sdk crashes when multiple resolves happen at the same time
+            # guard simultane resolves with a lock.
```
MartinHjelmare, line 863: simultaneous

matter_server/server/device_controller.py
MartinHjelmare: read_attributes

PR #387, opened 2023-09-08
Correct example for sending a command, format JSON examples

### LINKED ISSUE #384: Wrong example for sending a command
I find it hard to make sense of the example given for sending a command:

https://github.com/home-assistant-libs/python-matter-server/blob/dd6acaa819326a30ecb13c6069a7df6508a1edcc/README.md?plain=1#L270-L280

Are the message ID and command values correct? Endpoint and node ID values reference some missing code. Is the given dictionary meant to be a JSON example, outside the Python code?
How will the controller know which cluster command is being executed? In the given example, `payload` will be an empty object.

farshidtz:
Closes #384

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,102 +158,102 @@ Besides that it is possible to run Matter firmware on common microcontrollers su
 **Set WiFi credentials**
 Inform the controller about the WiFi credentials it needs to send when commissioning a new device.
 
-```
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
LGTM
Thanks!

PR #397, opened 2023-09-22
Fix scripts/generate_devices.py scripts

thsvkd:
Fix `gen_cls_name` function to generate the class name correctly even if `.` is in the device type name

I found `generate_devices.py` script has a bug when creating a class for a cluster corresponding to `PM2.5 Concentration Measurement` in `matter-devices.xml`. So, I modified the script code to ignore `.` in the device type name to sync with the connectedhomeip repository


marcelveldt:
Thanks!

PR #415, opened 2023-10-25
Cleanup/clarify commission_on_network

### LINKED ISSUE #408: setup_pin_code should be a string
The payload for on network commissioning expects the value of the setup_pin_code as a number/int
eg
```
{"message_id":"2","command":"commission_on_network","args":{"setup_pin_code":21036639901}}
```

However its possible and likely that the manual pairing code for a device will start with a leading zero but it will always be a fixed length. Many JSON libraries either refuse to allow a number value to start with a zero or they drop the leading zero.

Therefore it would be better to make this value a string to allow for codes that start with a zero

```
{"message_id":"2","command":"commission_on_network","args":{"setup_pin_code":"04970012338"}}
```

marcelveldt:
commission_on_network has a very confusing name and is actually not used for commissioning. I left it in the public api to not break anything but I've updated the docstrings and readme. commission_with_code is used for all scenarios, even for on-network devices like bridges and sharing from another controller. This was not very clear at this point.

fixes #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Inform the controller about the Thread credentials it needs to use when commissi
 ```
 
 **Commission with code**
-Commission a new device. For WiFi or Thread based devices, the credentials need to be set upfront, otherwise commisisoning will fail.
+Commission a new device. For WiFi or Thread based devices, the credentials need to be set upfront, otherwise commisisoning will fail. Supports both QR-code syntax (MT:...) and manual pairing code as string.
 The controller will use bluetooth for the commissioning of wireless devices. If the machine running the Python Matter Server controller lacks bluetooth support, commissioning will only work for devices already connected to the network (by cable or another controller).
 
 ```json
@@ -192,19 +192,6 @@ The controller will use bluetooth for the commissioning of wireless devices. If 
   "command": "commission_with_code",
   "args": {
     "code": "MT:Y.ABCDEFG123456789"
-  }
-}
-```
-
-**Commission on Network**
-Commission a device already connected to the network.
-
-```json
-{
-  "message_id": "2",
-  "command": "commission_on_network",
-  "args": {
-    "setup_pin_code": 1234567
   }
 }
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -183,7 +183,7 @@ Inform the controller about the Thread credentials it needs to use when commissi
 ```
 
 **Commission with code**
-Commission a new device. For WiFi or Thread based devices, the credentials need to be set upfront, otherwise commisisoning will fail.
+Commission a new device. For WiFi or Thread based devices, the credentials need to be set upfront, otherwise commisisoning will fail. Supports both QR-code syntax (MT:...) and manual pairing code as string.
```
MartinHjelmare, line 186:
```suggestion
Commission a new device. For WiFi or Thread based devices, the credentials need to be set upfront, otherwise commissioning will fail. Supports both QR-code syntax (MT:...) and manual pairing code as string.
```

matter_server/client/client.py
```diff
@@ -122,7 +122,13 @@ async def commission_with_code(self, code: str) -> MatterNodeData:
 
     async def commission_on_network(self, setup_pin_code: int) -> MatterNodeData:
         """
-        Commission a device already connected to the network.
+        Do the routine for OnNetworkCommissioning, with a filter for mDNS discovery.
+
+        The filter can be an integer,
```
MartinHjelmare, line 127: There's no filter in the parameters on the client side. It's not clear what we're talking about.

marcelveldt: I copied the docstring from the sdk but we shouldn't be using this method at all.
We should check our iOS and Android app if they're not using it and then just remove it. I was told this is only used for some of the CSA tests and all normal commissioning is just handled by the other command, which accepts all codes.

For now, I'll just adjust the docstring in the client to deprecated.

marcelveldt: updated the docstring

MartinHjelmare (approved):

PR #431, opened 2023-11-10
Remove special character

lboue:
This blocks CI tests


marcelveldt (approved):
oof, where did that come from? 
Thanks for spotting it!

PR #436, opened 2023-11-20
Add async-timeout

MartinHjelmare:
- Latest aiohttp now only includes async-timeout if running on Python 3.10 or lower.
- Include async-timeout explicitly for now until we bump minimum Python version. Python 3.11 has `asyncio.timeout`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ classifiers = [
 dependencies = [
   "aiohttp",
   "aiorun",
+  "async-timeout",
   "coloredlogs",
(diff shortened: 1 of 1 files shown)


frenck (approved):

agners:
Hm, I see that is why #435 failed. Should we maybe pin aiohttp? :thinking:

MartinHjelmare:
For libraries, which this is, we normally pin all requirements in development, but not in production. This library currently doesn't pin the production requirements in development.

PR #450, opened 2023-12-04
Implement more aggressive retry logic for resolving and first interview

marcelveldt:
Strictly taken this should not be needed but we've seen that thread networks can suffer from interference and/or discovery failing for some devices. This increases the chances that the commissioning of a node succeeds by simply retrying the resolving with extended timeouts.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -3,16 +3,17 @@
 from __future__ import annotations
 
 import asyncio
-from collections import deque
-from datetime import datetime
-from functools import partial
 import logging
 import random
 import time
+from collections import deque
+from datetime import datetime
+from functools import partial
 from typing import TYPE_CHECKING, Any, Callable, Iterable, TypeVar, cast
 
 from chip.ChipDeviceCtrl import CommissionableNode
-from chip.clusters import Attribute, Objects as Clusters
+from chip.clusters import Attribute
+from chip.clusters import Objects as Clusters
 from chip.clusters.Attribute import ValueDecodeFailure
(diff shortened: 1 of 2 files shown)


matter_server/server/device_controller.py
```diff
@@ -918,15 +935,22 @@ async def _resolve_node(self, node_id: int, retries: int = 3) -> None:
             async with node_lock, self._resolve_lock:
                 LOGGER.debug("Attempting to resolve node %s...", node_id)
                 await self._call_sdk(
-                    self.chip_controller.ResolveNode,
+                    self.chip_controller.GetConnectedDeviceSync,
```
MartinHjelmare, line 938: Is this the same as `ResolveNode`?

matter_server/server/device_controller.py
```diff
@@ -168,8 +168,23 @@ async def commission_with_code(self, code: str) -> MatterNodeData:
                 f"Commission with code failed for node {node_id}"
             )
 
-        # full interview of the device
-        await self.interview_node(node_id)
+        # perform full (first) interview of the device
+        # we retry the interview max 3 times as it may fail in noisy
+        # RF environments (in case of thread), mdns trouble or just flaky devices.
+        # retryuing both the mdns resolve and (first) interview, increases the chances
```
TheJulianJES, line 174: small typo in comment
```suggestion
        # retrying both the mdns resolve and (first) interview, increases the chances
```

matter_server/server/device_controller.py
MartinHjelmare: As discussed off line, it's the corresponding lower level call.

PR #458, opened 2023-12-07
Do not override the timeout of the operational node discovery

marcelveldt:
In another attempt of trying to fix the segfaulting sdk, remove the custom timeout on the "GetConnectedDeviceSync" as that seems to induce the issue (at least on my basic testing). Further research is needed to understand why this is happening but its very low level at the C++ side of things.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -932,14 +932,14 @@ class MatterDeviceController:
             # the sdk crashes when multiple resolves happen at the same time
             # guard simultane resolves with a lock.
             async with node_lock, self._resolve_lock:
-                LOGGER.debug("Attempting to resolve node %s...", node_id)
+                LOGGER.debug(
+                    "Attempting to resolve node %s... (attempt %s of %s)",
+                    node_id,
+                    attempt,
+                    retries,
+                )
                 await self._call_sdk(
-                    self.chip_controller.GetConnectedDeviceSync,
-                    allowPASE=False,
-                    # For the first attempts we use the SDK's default timeout.
-                    # Once we keep retrying we try the final attempt(s)
-                    # with an extended timeout.
-                    timeoutMs=30000 if attempt >= 3 else None,
+                    self.chip_controller.ResolveNode,
                     nodeid=node_id,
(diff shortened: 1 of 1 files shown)


agners:
From looking at `GetConnectedDeviceSync`, this timeout is not passed to the SDK, but is mearly usesd in the Python wrapper code to make the code return earlier (and ignoring any future callback). It seems that "ignoring future callback" is the issue here: When the Python code moves on, the callback function `DeviceAvailableCallback` gets GC'ed. This is a bug in the SDK. I've opened an issue at https://github.com/project-chip/connectedhomeip/issues/30870.

However, I think your fix here is the right thing to do: In our case using a timeout is not helpful. We want to know the *true* result of `GetConnectedDeviceSync`.

Sidenote: It seems that the SDK time out after about 45s naturally (added some debug prints in `GetConnectedDeviceSync`,  the id is just a local variable so I can match individual calls):
```
2023-12-07 14:35:10.022837 Calling for 547609483536 nodeid 8 with timeout None
2023-12-07 14:35:55.029104 DeviceAvailableCallback for 547609483536 with err src/lib/address_resolve/AddressResolve_DefaultImpl.cpp:119: CHIP Error 0x00000032: Timeout nodeid 8
2023-12-07 14:35:55.029624 Call done for 547609483536 nodeid 8
```

When using the timeout, the stack underneath continues to try to connect to Node 8. This is a trace with debug prints and a fixed SDK, the callback still comes back after 45s (even though we moved to node 12 here after 30s):
```
2023-12-07 14:37:40.068642 Calling for 547609065360 nodeid 8 with timeout 30000
2023-12-07 14:37:40.076642 Call done for 547609065360 nodeid 8
2023-12-07 14:38:10.078433 Calling for 547608698000 nodeid 12 with timeout 30000
2023-12-07 14:38:10.085811 Call done for 547608698000 nodeid 12
2023-12-07 14:38:25.079839 DeviceAvailableCallback for 547609065360 with err src/lib/address_resolve/AddressResolve_DefaultImpl.cpp:119: CHIP Error 0x00000032: Timeout nodeid 8
```

PR #462, opened 2023-12-13
Log extensively during commissioning

agners:
This helps users in debugging and better understanding at what stage commissioning fails. Commissioning is not happening all that often, so logging extensively shouldn't hurt in this case.


agners:
I moved `parse_attributes_from_read_result()` since pylint started to complain about length of the device_controller module (see https://github.com/home-assistant-libs/python-matter-server/actions/runs/7197971979/job/19606327910).

matter_server/server/device_controller.py
```diff
@@ -155,18 +157,24 @@ async def commission_with_code(self, code: str) -> MatterNodeData:
 
         # perform a quick delta sync of certificates to make sure
         # we have the latest paa root certs
+        LOGGER.info("Commissioning new device with code.")
```
marcelveldt, line 160: this is duplicate

matter_server/server/device_controller.py
```diff
@@ -176,17 +184,22 @@ async def commission_with_code(self, code: str) -> MatterNodeData:
         retries = 3
         while retries:
             try:
+                LOGGER.info(
```
marcelveldt, line 187: why? this is also logged in the interview method

matter_server/server/device_controller.py
```diff
@@ -215,10 +228,14 @@ async def commission_on_network(
         # we have the latest paa root certs
         # NOTE: Its not very clear if the newly fetched certificates can be used without
         # restarting the device controller
+        LOGGER.info("Commissioning new device with code.")
```
marcelveldt, line 231: duplicate

matter_server/server/device_controller.py
```diff
@@ -215,10 +228,14 @@ async def commission_on_network(
...
         await fetch_certificates()
```
marcelveldt, line 232: this will be removed in a patch I have waiting

matter_server/server/device_controller.py
```diff
@@ -227,14 +244,18 @@ async def commission_on_network(
             filter=filter,
         )
         if not success:
+            LOGGER.warning("Commission on network failed for node %s", node_id)
```
marcelveldt, line 247: Why this log, the exception will be thrown as well

matter_server/server/device_controller.py
```diff
@@ -155,18 +157,24 @@ async def commission_with_code(self, code: str) -> MatterNodeData:
...
+        LOGGER.info(
+            "Starting Matter commissioning with code using Node ID %s.", node_id
+        )
         success = await self._call_sdk(
             self.chip_controller.CommissionWithCode,
             setupPayload=code,
             nodeid=node_id,
         )
         if not success:
+            LOGGER.warning("Commission with code failed for node %s", node_id)
```
marcelveldt, line 173: why? the exception will be raised and printed

matter_server/server/device_controller.py
```diff
@@ -227,14 +244,18 @@ async def commission_on_network(
...
         )
         if not success:
+            LOGGER.warning("Commission on network failed for node %s", node_id)
             raise NodeCommissionFailed(
                 f"Commission on network failed for node {node_id}"
             )
+        LOGGER.info("Matter commissioning of Node ID %s successful.", node_id)
 
         # full interview of the device
+        LOGGER.info("Start interview of newly commissioned Node ID %s.", node_id)
```
marcelveldt, line 254: duplicate

matter_server/server/device_controller.py
```diff
@@ -897,26 +918,6 @@ def reschedule() -> None:
             # https://github.com/project-chip/connectedhomeip/pull/26718
             reschedule()
 
-    @staticmethod
```
marcelveldt, line 900: this has nothing to do with logging while I do agree that it could be moved to a helper module.
Maybe update the PR description ?

marcelveldt:
> I moved `parse_attributes_from_read_result()` since pylint started to complain about length of the device_controller module (see https://github.com/home-assistant-libs/python-matter-server/actions/runs/7197971979/job/19606327910).

ah I see this now after the review. All clear

matter_server/server/device_controller.py
agners: It is before fetching the certificate... What if certificate fetching is slow, and causes the commissioning to break? I think it is really beneficial to see what is happening during commissioning to pinpoint problems.

If we drop the certificate fetching, then I am ok with removing this.

matter_server/server/device_controller.py
agners: I see. But in debug level only. I think this doesn't happen all to often, so loggin in info seems sensible to me. Should we raise in `interview_node`?

matter_server/server/device_controller.py
marcelveldt: Yeah, my local branch with a bunch of patches is almost ready which drops this certificate fetching (as it doesnt work anyways, only at startup)

matter_server/server/device_controller.py
marcelveldt: yes please, that makes more sense

marcelveldt (approved):

Issue #463, opened 2023-12-13
"Something went wrong" when commissioning through Android app


agners:
When trying to commission a device through the Android App sometimes simply "Something went wrong" is returned. 

<img src="https://github.com/home-assistant-libs/python-matter-server/assets/34061/64d3c7e6-2235-41cd-8a1c-bec5a245d48a" width=50% height=50%>

Note: There might be multiple issue behind that particular error, but this outlines what I've been able to reproduce here.

There are two ways to on-board a Matter device to Home Assistant using Android:

1. Waiting for the half-screen pop-up to appear, then choose the Home Assistant app
2. Open the Home Assistant App, then adding a Matter device  by navigating to Settings > Devices & services > Devices > Add Device > Add Matter device

When using method 2, the Android app uses the `matter/commission_on_network` WS command is used. The passcode is passed to this endpoint, no further information. The Matter Server (SDK) then searches for *any* commissionable device on the network, and (presumably) picks the first it hears about. This of course could be a different one.

Furthermore, a BR might cache DNS-SD entry for a commissionable device (any BR runs a SRP server, which caches service announcements on behalf of Thread devices). Now if a Thread device got reset during commissioning at one point, such a stale entry might linger around for up to 2h (in my particular test case).

I've noticed this by checking the service announcements during a failed commissioning attempt. Two announcements were present:

```
$ sudo dns-sd -B _matterc._udp local
Browsing for _matterc._udp.local
DATE: ---Wed 13 Dec 2023---
19:45:23.143  ...STARTING...
Timestamp     A/R    Flags  if Domain               Service Type         Instance Name
19:45:23.143  Add        3   2 local.               _matterc._udp.       14C551B182BE1DB4
19:45:23.143  Add        2   2 local.               _matterc._udp.       9FD5C3686D913371  
```

And it turned out that in this case the second announcement was from a stale Thread device from the OTBR Add-on:

(I've run this command for debugging from within the container)
```
$ ot-ctl srp server service
...
9FD5C3686D913371._matterc._udp.default.service.arpa.
    deleted: false
    subtypes: _V4447,_T65535,_S15,_L3840
    port: 5540
    priority: 0
    weight: 0
    ttl: 7200
    lease: 7200
    key-lease: 680400
[...]

PR #469, opened 2023-12-18
Add IP address support for commission_on_network

agners:
When using the Android Matter commissioning flow from the Home Assistant App (via CommissioningRequest), we get the passcode only, no QR code or manual pairing code (see CommissioningRequestMetadata documentation). This means there is no discriminator. In other words, the server will try to find a commissionable device, and use the pincode against the first one found.

It seems that sometimes another device is announced to be commissionable at the same time, which obvisouly has another passcode, hence the commission fails quite quickly. The App shows "Something went wrong".

This seems to be particularly common when using Thread: The Thread border router forwards and caches DNS-SD/mDNS service information about commissionable Thread devices with its SRP server service. The cache is up to 2h long. So when a communication breakdown happens after a particular device went into commission mode, this commissionable service entry lingers in the network for quite some time.

Now if the OTBR sends this entry before the new/valid entry, the Matter Server tries to commission a device which is no longer responding.

The CommissioningRequestMetadata have a way to get the device's IP address. This change extends commission_on_network WS endpoint to also take an IP address. The SDK CommissionIP service is used to commission the device.

Note: CommissionIP is marked deprecated currently. This is mainly to prevent implementation which would ask users for IP addresses, which is not the intended way to implement commissioning. However, for this particular usecase the API seems very sensible and works as intended.

Fixes: #463

### DIFF
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -120,7 +120,7 @@ class MatterClient:
         data = await self.send_command(APICommand.COMMISSION_WITH_CODE, code=code)
         return dataclass_from_dict(MatterNodeData, data)
 
-    async def commission_on_network(self, setup_pin_code: int) -> MatterNodeData:
+    async def commission_on_network(self, setup_pin_code: int, ip_addr: str | None = None) -> MatterNodeData:
         """
         Do the routine for OnNetworkCommissioning.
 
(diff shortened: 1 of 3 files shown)


MartinHjelmare (approved):

satmandu:
Really looking forward to seeing if this solves my commissioning issues with my Onvis S4 devices!

PR #470, opened 2023-12-21
Allow to commission with code on network only

agners:
Add a parameter which instructs the SDK to search for the device only on the network. This is useful when using the manual pairing code, which does not contain a device specific discovery capabilties.

This is useful to prevent the SDK from searching via Bluetooth from the Android App (which provides a manual pairing code when using the non-App integrated Matter onboarding flow).

### DIFF
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -112,7 +112,7 @@ class MatterClient:
         raise NodeNotExists(f"Node {node_id} does not exist or is not yet interviewed")
 
     async def commission_with_code(
-        self, code: str, network_only: bool
+        self, code: str, network_only: bool = False
     ) -> MatterNodeData:
         """
         Commission a device using QRCode or ManualPairingCode.
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -115,9 +115,12 @@ class MatterClient:
         self, code: str, network_only: bool = False
     ) -> MatterNodeData:
         """
-        Commission a device using QRCode or ManualPairingCode.
+        Commission a device using a QR Code or Manual Pairing Code.
 
-        Returns basic MatterNodeData once complete.
+        :param code: The QR Code or Manual Pairing Code for device commissioning.
+        :param network_only: If True, restricts device discovery to network only.
+
+        :return: The NodeInfo of the commissioned device.
         """
         data = await self.send_command(
             APICommand.COMMISSION_WITH_CODE, code=code, network_only=network_only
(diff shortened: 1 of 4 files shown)


matter_server/client/client.py
```diff
@@ -111,13 +111,17 @@ def get_node(self, node_id: int) -> MatterNode:
             return node
         raise NodeNotExists(f"Node {node_id} does not exist or is not yet interviewed")
 
-    async def commission_with_code(self, code: str) -> MatterNodeData:
+    async def commission_with_code(
+        self, code: str, network_only: bool
```
MartinHjelmare, line 115: Should `network_only` be required parameter or should we set a default like in the server command?

agners: Good point, let's make it an optional parameter as well :+1:

matter_server/server/device_controller.py
```diff
@@ -1,4 +1,5 @@
 """Controller that Manages Matter devices."""
+# pylint: disable=C0302
```
marcelveldt, line 2: can we use the human-readable name here instead of the code ?

matter_server/client/client.py
```diff
@@ -111,13 +111,17 @@ def get_node(self, node_id: int) -> MatterNode:
...
         raise NodeNotExists(f"Node {node_id} does not exist or is not yet interviewed")
 
-    async def commission_with_code(self, code: str) -> MatterNodeData:
+    async def commission_with_code(
+        self, code: str, network_only: bool = False
+    ) -> MatterNodeData:
         """
         Commission a device using QRCode or ManualPairingCode.
 
         Returns basic MatterNodeData once complete.
```
marcelveldt, line 120: should we add a note about the new param in the docstring for lib consumers ?

matter_server/server/device_controller.py
```diff
@@ -145,7 +146,9 @@ def get_node(self, node_id: int) -> MatterNodeData:
         raise NodeNotExists(f"Node {node_id} does not exist or is not yet interviewed")
 
     @api_command(APICommand.COMMISSION_WITH_CODE)
-    async def commission_with_code(self, code: str) -> MatterNodeData:
+    async def commission_with_code(
+        self, code: str, network_only: bool = False
+    ) -> MatterNodeData:
         """
         Commission a device using QRCode or ManualPairingCode.
```
marcelveldt, line 153: should we add a note about the new param in the docstring  ?

MartinHjelmare (approved):

PR #471, opened 2023-12-21
Optimize commission and (re)subscription flows

marcelveldt:
A bunch of (small) optimizations in the commissioning and (re)subscription flows again based on some investigation with various devices. 

- Remove fetching certificates before commissioning (as it only slows down and doesn't do anything)
- Ignore failed commissioned nodes at startup
- Cancel subscription at node removal
- Speed up resolve
- Split up node attribute subscriptions into 2 subscriptions
- Lower the ceiling once again
- Adjust logging of PersistentStorage module of sdk because it was too chatty

### DIFF
--- a/matter_server/server/__main__.py
+++ b/matter_server/server/__main__.py
@@ -72,6 +72,7 @@ def main() -> None:
     coloredlogs.install(level=args.log_level.upper())
     if not logging.getLogger().isEnabledFor(logging.DEBUG):
         logging.getLogger("chip").setLevel(logging.WARNING)
+        logging.getLogger("PersistentStorage").setLevel(logging.WARNING)
 
     # make sure storage path exists
     if not os.path.isdir(args.storage_path):
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -153,9 +153,6 @@ class MatterDeviceController:
         if self.chip_controller is None:
             raise RuntimeError("Device Controller not initialized.")
 
-        # perform a quick delta sync of certificates to make sure
-        # we have the latest paa root certs
-        await fetch_certificates()
         node_id = self._get_next_node_id()
 
         success = await self._call_sdk(
@@ -210,12 +207,6 @@ class MatterDeviceController:
         """
         if self.chip_controller is None:
             raise RuntimeError("Device Controller not initialized.")
-
-        # perform a quick delta sync of certificates to make sure
-        # we have the latest paa root certs
-        # NOTE: Its not very clear if the newly fetched certificates can be used without
-        # restarting the device controller
-        await fetch_certificates()
 
         node_id = self._get_next_node_id()
 
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -468,6 +468,7 @@ class MatterDeviceController:
             DATA_KEY_NODES,
             subkey=str(node_id),
         )
+        self.server.storage.save(immediate=True)
 
         assert node is not None
 
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -444,7 +444,6 @@ class MatterDeviceController:
         await self._resolve_node(node_id=node_id)
         endpoint_id, cluster_id, attribute_id = parse_attribute_path(attribute_path)
         async with node_lock:
-            self.chip_controller.CheckIsActive()
             assert self.server.loop is not None
             future = self.server.loop.create_future()
             device = await self._resolve_node(node_id)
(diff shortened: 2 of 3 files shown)


matter_server/server/__main__.py
```diff
@@ -72,6 +72,7 @@ def main() -> None:
     coloredlogs.install(level=args.log_level.upper())
     if not logging.getLogger().isEnabledFor(logging.DEBUG):
         logging.getLogger("chip").setLevel(logging.WARNING)
+        logging.getLogger("PersistentStorage").setLevel(logging.WARNING)
```
agners, line 75: This is a rather unrelated change and should be a separate commit (that also has the advantage that it appears in changelogs etc.)

matter_server/server/device_controller.py
```diff
@@ -99,12 +109,16 @@ async def initialize(self) -> None:
...
+        nodes: dict[str, dict | None] = self.server.storage.get(DATA_KEY_NODES, {})
         for node_id_str, node_dict in nodes.items():
             node_id = int(node_id_str)
-            # invalidate node data if schema mismatch,
-            # the node will automatically be scheduled for re-interview
-            if node_dict and node_dict.get("interview_version") != SCHEMA_VERSION:
+            if node_dict is None:
+                # ignore non-initialized (left-over) nodes
+                # from failed commissioning attempts
+                continue
```
agners, line 118: IMHO we should still retry to interview those. It could be that the user commissioned them, then interviewing failed, he restarts the add-on and now maybe this second time interviewing works... :man_shrugging: 

I understand that now people have probably such stale entries, and this code makes it retry every time. And since people don't see those in Core, they also cant' delete them.

So, as a work around, I think this is fine.

But once we should them, have them cleaned-up, we should no longer ignore them.

matter_server/server/device_controller.py
```diff
@@ -630,59 +655,83 @@ async def _subscribe_node(self, node_id: int) -> None:
...
+        """Setup a single Node AttributePath(s) subscription."""
+        node_id = node.node_id
+        node_logger = LOGGER.getChild(f"[node {node_id}]")
+        assert self.chip_controller is not None
+        node_logger.debug("Setting up attributes and events subscription.")
+        self._last_subscription_attempt[node_id] = 0
+        self.chip_controller.CheckIsActive()
+        assert self.server.loop is not None
+        future = self.server.loop.create_future()
+        device = self.chip_controller.GetConnectedDeviceSync(node_id)
```
agners, line 717: Do we need the resolve lock here (since in the comment in `_resolve_node()` it says simultaneous resolving causes SDK crashes).

matter_server/server/device_controller.py
marcelveldt: No, we agreed on this. This is old stuff. It can not happen anymore but we need to clean out the old traces.

matter_server/server/device_controller.py
marcelveldt: Oof, yes, good catch!

matter_server/server/__main__.py
marcelveldt: Yes it should be a seperated commit indeed, I'll put it in a seperate commit

marcelveldt: but well in the end this will be a squash commit so it doesnt matter after all. I'll add it to the PR desciption

matter_server/server/device_controller.py

matter_server/server/device_controller.py
```diff
@@ -630,59 +658,82 @@ async def _subscribe_node(self, node_id: int) -> None:
...
+        interval_ceiling: int = 60,
+        event_subscriptions: list[Attribute.EventPath] | None = None,
+    ) -> Attribute.SubscriptionTransaction:
+        """Handle Setup of a single Node AttributePath(s) subscription."""
+        node_id = node.node_id
+        node_logger = LOGGER.getChild(f"[node {node_id}]")
+        assert self.chip_controller is not None
+        node_logger.debug("Setting up attributes and events subscription.")
+        self._last_subscription_attempt[node_id] = 0
+        assert self.server.loop is not None
```
MartinHjelmare, line 717: We're using the loop attribute multiple times. I'd store it in a local variable first.

MartinHjelmare (approved):

matter_server/server/device_controller.py
```diff
@@ -131,8 +131,21 @@ async def start(self) -> None:
...
+                # if the node was not available last time we will delay
+                # the first attempt to initialize so that we prioritize nodes
+                # that are probably available so they are back online as soon as
+                # possible and we're not stuck trying to initialize nodes that are offline
+                assert self.server.loop
+                self.server.loop.call_later(
+                    5,
+                    asyncio.create_task,
+                    self._check_interview_and_subscription(node_id),
+                )
```
MartinHjelmare, line 144: It's better to have a callback that creates the task so that we don't have a non awaited coroutine possibly lingering if the server is shutdown before the delay.

PR #472, opened 2023-12-22
Move delayed interview to own callback

marcelveldt:
Prevent lingering task errors on early shutdown by moving the scheduled interview to its own function.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -954,12 +954,16 @@ class MatterDeviceController:
         assert self.server.loop is not None
         # pop any existing (re)schedule timer
         self._sub_retry_timer.pop(node_id, None)
+
+        def create_interview_task() -> None:
+            asyncio.create_task(
+                self._check_interview_and_subscription(
+                    node_id,
+                )
+            )
+
         self._sub_retry_timer[node_id] = self.server.loop.call_later(
-            delay,
-            asyncio.create_task,
-            self._check_interview_and_subscription(
-                node_id,
-            ),
+            delay, create_interview_task
         )
 
     async def _resolve_node(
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -962,7 +944,23 @@ def reschedule() -> None:
...
+
+    def _schedule_interview(self, node_id: int, delay: int) -> None:
+        """(Re)Schedule interview and/or initial subscription for a node."""
+        assert self.server.loop is not None
+        # pop any existing (re)schedule timer
+        self._sub_retry_timer.pop(node_id, None)
+        self._sub_retry_timer[node_id] = self.server.loop.call_later(
+            delay,
+            asyncio.create_task,
+            self._check_interview_and_subscription(
```
MartinHjelmare, line 960: We create the coroutine here and then wait delay until the task is created. Instead we should wait delay and then create the coroutine and schedule it as a task. Make an additional callback that creates the task and we can use as target for call_later.

MartinHjelmare (changes requested):

matter_server/server/device_controller.py
marcelveldt: Yes, I need more coffee

MartinHjelmare (approved):
Thanks!

matter_server/server/device_controller.py
```diff
@@ -962,7 +944,27 @@ def reschedule() -> None:
...
+            self._schedule_interview(
+                node_id,
+                min(reschedule_interval + 10, MAX_POLL_INTERVAL),
+            )
+
+    def _schedule_interview(self, node_id: int, delay: int) -> None:
+        """(Re)Schedule interview and/or initial subscription for a node."""
+        assert self.server.loop is not None
+        # pop any existing (re)schedule timer
+        self._sub_retry_timer.pop(node_id, None)
```
MartinHjelmare, line 956: Should we cancel the call_later task too?

MartinHjelmare (commented):
One question.

PR #480, opened 2023-12-27
Disable chip.Attribute logger

marcelveldt:
temporary disable the logger of chip.clusters.Attribute because it now logs an error on every custom attribute that couldn't be parsed which confuses people.

We can restore the default log level again when we patched the device controller to handle the raw attribute data to deal with custom clusters.

### DIFF
--- a/matter_server/server/__main__.py
+++ b/matter_server/server/__main__.py
@@ -73,6 +73,11 @@ def main() -> None:
     if not logging.getLogger().isEnabledFor(logging.DEBUG):
         logging.getLogger("chip").setLevel(logging.WARNING)
         logging.getLogger("PersistentStorage").setLevel(logging.WARNING)
+        # temporary disable the logger of chip.clusters.Attribute because it now logs
+        # an error on every custom attribute that couldn't be parsed which confuses people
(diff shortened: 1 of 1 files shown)


matter_server/server/__main__.py
```diff
@@ -73,6 +73,11 @@ def main() -> None:
     if not logging.getLogger().isEnabledFor(logging.DEBUG):
         logging.getLogger("chip").setLevel(logging.WARNING)
         logging.getLogger("PersistentStorage").setLevel(logging.WARNING)
+        # temporary disable the logger of chip.clusters.Attribute because it now logs
+        # an error on every custom attribute that couldn't be parsed which confuses people
+        # we can restore the default log level again when we patched the device controller
+        # to handle the raw attribute data to deal with custom clusters.
```
MartinHjelmare, line 79:
```suggestion
        # Temporary disable the logger of chip.clusters.Attribute because it now logs
        # an error on every custom attribute that couldn't be parsed which confuses people.
        # We can restore the default log level again when we've patched the device controller
        # to handle the raw attribute data to deal with custom clusters.
```

MartinHjelmare (approved):

PR #481, opened 2023-12-27
Restore single subscription to a node instead of dual

### LINKED ISSUE #476: Reliability regression from 5.0.1 to stable
Hey,
this one took a while to figure out.

I've tried matter together with a ESP32 example firmware (so just WiFi) as an UnRAID Docker.
While everything worked fine when adding the device to Alexa and there were absolutely no reliability issues, when adding it to Home-Assistant using the latest stable version I get a super unstable experience after just a couple of seconds. Because of the finnicky IPV6 setup I've first started trying different Hosts, so I've spun up the docker on multiple other host machines (Arch VM, UbuntuServer VM, Raspberry Pi 3 running Raspbian 12) with the same outcome. 

This was with current stable and the version before that.

It was when I tried HAOS on a VM when it started working and I thought it had something to do with kernel/networkmanager patches in this distribution. Turns out the HAOS addon was still using an older version 5.0.1. 

So I've tried this version on one of the other hosts and it worked! Guess there has been a regression somewhere in between?! I'd love to help troubleshoot this.

Logs with current stable:
https://pastebin.com/EAFpB0z4

Logs with 5.0.1
https://pastebin.com/mVy5g8rU

marcelveldt:
A dual subscription to the same node appears to be more unstable than maintaining a single subscription with a maximum of 9 paths. Restore our code so we only handle one single sub per node instead of multiple, while keeping the other fixes we did lately.

Fixes #476 
Fixes https://github.com/home-assistant/core/issues/105480

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -82,10 +82,7 @@ class MatterDeviceController:
         self.server = server
         # we keep the last events in memory so we can include them in the diagnostics dump
         self.event_history: deque[Attribute.EventReadResult] = deque(maxlen=25)
-        self._subscriptions: dict[
-            int,
-            tuple[Attribute.SubscriptionTransaction, Attribute.SubscriptionTransaction],
-        ] = {}
+        self._subscriptions: dict[int, Attribute.SubscriptionTransaction] = {}
         self._attr_subscriptions: dict[int, list[Attribute.AttributePath]] = {}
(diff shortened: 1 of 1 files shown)


marcelveldt:
Tested this for about an hour and the devices stay connected and are not constantly resubscribing anymore.

frenck (approved):

PR #488, opened 2024-01-04
Add retries to all commissioning paths

marcelveldt:
Not all commissioning paths had some retries applied. Now we retry both commisisoning and first interview 3 times on both the code and on network commissisioning.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -66,7 +66,7 @@ BASE_SUBSCRIBE_ATTRIBUTES: tuple[Attribute.AttributePath, Attribute.AttributePat
     ),
 )
 
-# pylint: disable=too-many-lines,too-many-locals,too-many-statements
+# pylint: disable=too-many-lines,too-many-locals,too-many-statements,too-many-branches
 
 
 class MatterDeviceController:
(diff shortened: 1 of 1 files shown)


agners:
The main goal is to resolve commissioning issues such as observed with Meross MSS315 plugs. It seems that they fail at certain commissioning step, e.g. in this case 18:

```
2024-01-04 14:55:51 core-matter-server matter_server.server.device_controller[212] INFO Starting Matter commissioning with code using Node ID 16.
2024-01-04 14:56:27 core-matter-server chip.CTL[212] ERROR Failed to perform commissioning step 18
2024-01-04 14:56:27 core-matter-server matter_server.server.client_handler[212] ERROR [549496677840] Error handling message: CommandMessage(message_id='85292457bd0746e09707e81ae14e8709', command='commission_with_code', args={'code': '10613221685', 'network_only': True})
  File "/usr/local/lib/python3.11/site-packages/matter_server/server/client_handler.py", line 188, in _run_handler
    result = await result
             ^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/matter_server/server/device_controller.py", line 196, in commission_with_code
    raise NodeCommissionFailed(
matter_server.common.errors.NodeCommissionFailed: Commission with code failed for node 16
2024-01-04 14:56:27 core-matter-server chip.EM[212] ERROR Failed to send Solitary ack for MessageCounter:7068048 on exchange 59541i:src/messaging/ExchangeContext.cpp:103: CHIP Error 0x00000002: Connection aborted
```

According to the SDK this step is the step where the NOC is added to the device (see https://github.com/project-chip/connectedhomeip/blob/v1.2.0.1/src/controller/CommissioningDelegate.h#L33-L72, `kSendNOC`).

There isn't much we can do if a device just disappears during commissioning other then retrying :cry:

matter_server/server/device_controller.py
```diff
@@ -186,16 +186,22 @@ async def commission_with_code(
...
+                networkOnly=network_only,
             )
+            if success:
+                break
+            if not success and not retries:
+                raise NodeCommissionFailed(
+                    f"Commission with code failed for node {node_id}"
+                )
+            await asyncio.sleep(5)
+            retries -= 1
```
agners, line 204: Can you add a message here that we are retrying? It will be helpful to understand logs better when we know that multiple commissioning attempts are being made for a particular device.

marcelveldt:
Yeah this is a workaround for what appears to be a device (or maybe other device have this as well?) that is just a bit unstable. By adding some retries we increase the chances of a successful commissioning. May also help with the famous Nanoleaf devices to get them onboarded.

matter_server/server/device_controller.py
marcelveldt: I have adjusted the logging, now we will log the attempt too

PR #448, opened 2023-12-04
Update json dump helper to handle CommissionableNode structure

oidebrett:
TypeError: Type is not JSON serializable: CommissionableNode

python3 --version Python 3.10.12

Suspected Reason:
json dumps handler cannot handle CommissionableNode structure

Steps to reproduce:

1) flash latest esp32 lighting-app example to ESP32 dev kit

2) commission esp32 lighting-app example using chip-tool
./chip-tool pairing ble-wifi 0x1 MYSSID MYPASS 20202021 3840

3) Once commissioning is successful as node 1, open a commissioning window:
./chip-tool pairing open-commissioning-window 1 1 300 1000 2365

4) run a discover from the python-matter-server
{"message_id": "1","command": "discover"}

5) python-matter-server log shows:
023-12-04 12:00:57 dell matter_server.server.client_handler[2017833] DEBUG [140135655073504] Received CommandMessage(message_id='1', command='discover', args=None)
2023-12-04 12:00:57 dell matter_server.server.client_handler[2017833] DEBUG [140135655073504] Handling command discover
2023-12-04 12:00:57 dell chip.DL[2017833] INFO Avahi browse: cache new
2023-12-04 12:00:57 dell chip.DL[2017833] INFO Avahi browse: cache new
2023-12-04 12:00:57 dell chip.DL[2017833] INFO Avahi browse: cache exhausted
2023-12-04 12:00:58 dell chip.DL[2017833] INFO Avahi browse: all for now
2023-12-04 12:00:58 dell chip.DL[2017833] INFO Avahi resolve found
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG Discovered node:
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG        Hostname: 58BF2593B9A8
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG        IP Address #1: 192.168.1.41
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG        Port: 5540
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG        Mrp Interval idle: not present
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG        Mrp Interval active: not present
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG        TCP Supported: 1
2023-12-04 12:00:58 dell chip.DIS[2017833] DEBUG        Vendor ID: 65521
[...]

PR #490, opened 2024-01-05
Fix json serialization for discovery api

wozz:
alternative to #448 

validated locally:

```json
{
  "message_id": "c9627ca8-b91d-4f76-98de-91d4f34cd495",
  "result": [
    {
      "instanceName": "<redacted>",
      "hostName": "<redacted>",
      "port": <redacted>,
      "longDiscriminator": <redacted>,
      "vendorId": 4874,
      "productId": 89,
      "commissioningMode": 2,
      "deviceType": 263,
      "deviceName": "Eve Motion",
      "pairingInstruction": "",
      "pairingHint": <redacted>,
      "mrpRetryIntervalIdle": 3300,
      "mrpRetryIntervalActive": 1100,
      "supportsTcp": false,
      "addresses": [
        "<redacted>"
      ],
      "rotatingId": "<redacted>"
    }
  ]
}
```

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -16,6 +16,7 @@ from chip.ChipDeviceCtrl import (
     CommissioningParameters,
     DeviceProxyWrapper,
 )
+from chip.discovery import CommissionableNode as CommissionableNodeData
 from chip.clusters import Attribute, Objects as Clusters
 from chip.clusters.Attribute import ValueDecodeFailure
 from chip.clusters.ClusterObjects import ALL_ATTRIBUTES, ALL_CLUSTERS, Cluster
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
Exactly what was needed, thanks

marcelveldt:
CI reports an (linting?) issue

matter_server/server/device_controller.py
```diff
@@ -383,15 +384,24 @@ async def open_commissioning_window(
...
             raise RuntimeError("Device Controller not initialized.")
 
         result = await self._call_sdk(
             self.chip_controller.DiscoverCommissionableNodes,
         )
-        return result
+
+        def convert(cn: CommissionableNode) -> CommissionableNodeData:
+            cnd = CommissionableNodeData()
+            for field in CommissionableNodeData.__dataclass_fields__:
```
wozz, line 398: linter reports:

```
matter_server/server/device_controller.py:398:25: E1101: Class 'CommissionableNode' has no '__dataclass_fields__' member (no-member)
```

I'm not quite sure the issue - could be the linter is confused about `CommissionableNode` from `chip.ChipDeviceCtrl` vs `chip.discovery` - however since it's a subclass, I would expect it to have all the same fields available. Could be that `__dataclass_fields__` field is created in some way that the linter is unable to detect.

Should I just ignore the lint warning here? Or are there any ideas to fix it?

marcelveldt: Yeah, its because they subclassed with a completely different signature and even with the same name.
Let's ignore for now yeah, I don't have any better options.

PR #500, opened 2024-01-11
Print result of device removal

agners:
Ideally, we like to remove the Home Assistant fabric from a device. We optimistically execute the command to remove the device.

Check the result of the device removal and print a message accordingly so users can tell if removing was successful.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -576,6 +576,8 @@ class MatterDeviceController:
         self.server.storage.save(immediate=True)
         LOGGER.info("Node ID %s successfully removed from Matter server.", node_id)
 
+        self.server.signal_event(EventType.NODE_REMOVED, node_id)
+
         assert node is not None
 
         attribute_path = create_attribute_path_from_attribute(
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -584,15 +586,25 @@ async def remove_node(self, node_id: int) -> None:
...
+            )
         )
+        if (
+            result.statusCode
+            == Clusters.OperationalCredentials.Enums.NodeOperationalCertStatusEnum.kOk
+        ):
+            LOGGER.info("Successfully removed Home Assistant fabric from device.")
+        else:
+            LOGGER.warning(
+                "Removing Home Assistant fabric from device failed with status code %d.",
```
marcelveldt, line 605: strictly taken this is not the home assistant fabric but the fabric from the consumer of our lib.

marcelveldt, line 605:
```suggestion
                "Removing current fabric from device failed with status code %d.",
```

marcelveldt (approved):
left one suggestion, otherwise lgtm

matter_server/server/device_controller.py
agners: Yeah good point :+1:  Thanks!

PR #501, opened 2024-01-11
Scope link-local IPv6 addresses received from WebSocket

agners:
Scope link-local IPv6 addresses received from WebSocket

When receiving link-local IPv6 addresses, remove the scope by default
as the scope id is typically from a remote machine (e.g. from the Android
phone).

If the Matter server got started with a primary interface set, scope the
link-local address to that interface instead.

This fixes commissioning issue for WiFi devices when using the Android
in-app commissioning flow which sends the IP address of the device
to be commissioned alongside.

Related-to: #463

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -259,6 +259,8 @@ class MatterDeviceController:
             raise RuntimeError("Device Controller not initialized.")
 
         node_id = self._get_next_node_id()
+        if ip_addr is not None:
+            ip_addr = self.server.scope_ipv6_lla(ip_addr)
 
         attempts = 0
         # we retry commissioning a few times as we've seen devices in the wild
@@ -281,9 +283,8 @@ class MatterDeviceController:
                     filter=filter,
                 )
             else:
-                ip_addr = self.server.scope_ipv6_lla(ip_addr)
                 LOGGER.info(
-                    "Starting Matter commissioning with IP using Node ID %s via IP address %s  (attempt %s/%s).",
+                    "Starting Matter commissioning using Node ID %s and IP %s (attempt %s/%s).",
                     node_id,
                     ip_addr,
                     attempts,
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -281,10 +281,11 @@ class MatterDeviceController:
                     filter=filter,
                 )
             else:
-                # Ip-adress provided, use CommissionIP method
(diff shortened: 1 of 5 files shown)


agners:
I've tested that with

matter_server/server/device_controller.py
```diff
@@ -281,10 +281,11 @@ async def commission_on_network(
                     filter=filter,
                 )
             else:
-                # Ip-adress provided, use CommissionIP method
+                ip_addr = self.server.scope_ipv6_lla(ip_addr)
                 LOGGER.info(
-                    "Starting Matter commissioning with IP using Node ID %s (attempt %s/%s).",
+                    "Starting Matter commissioning with IP using Node ID %s via IP address %s  (attempt %s/%s).",
```
marcelveldt, line 286:
```suggestion
                    "Starting Matter commissioning with IP "
                    "using Node ID %s via IP address %s  (attempt %s/%s).",
```

marcelveldt (approved):
left one comment (to fix the lint issue)

matter_server/server/device_controller.py
agners: I've just updated that part, made the message bit shorter.

Jc2k (approved):

matter_server/server/server.py
```diff
@@ -166,6 +169,34 @@ def signal_event(self, evt: EventType, data: Any = None) -> None:
...
+        # Rely on host OS routing table
+        if self.primary_interface is None:
+            return ip_addr
+
+        self.logger.debug(
+            "Setting scope of link-local IP address %s to %s",
+            ip_addr,
+            self.primary_interface,
+        )
+        return ip_addr + "%" + self.primary_interface
```
MartinHjelmare, line 198: Please use an f-string instead of string concatenation.

pyproject.toml
```diff
@@ -28,12 +28,12 @@ dependencies = [
   "coloredlogs",
   "dacite",
   "orjson",
-  "home-assistant-chip-clusters==2023.12.0"
+  "home-assistant-chip-clusters==2024.1.0"
```
MartinHjelmare, line 31: The wheels bump wasn't mentioned in the PR description.

agners: Right, I've missed mentioning that :cry: 

Added it to the PR description!

matter_server/server/server.py
agners: Addressed with #504.

PR #503, opened 2024-01-15
Fix Node removal

marcelveldt:
There was a small oversight in the recently added code to log the success of a node removal.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -587,8 +587,9 @@ class MatterDeviceController:
             Clusters.OperationalCredentials.Attributes.CurrentFabricIndex,
         )
         fabric_index = node.attributes[attribute_path]
+        result: Clusters.OperationalCredentials.Commands.NOCResponse | None = None
         try:
-            await self.chip_controller.SendCommand(
+            result = await self.chip_controller.SendCommand(
                 nodeid=node_id,
                 endpoint=0,
                 payload=Clusters.OperationalCredentials.Commands.RemoveFabric(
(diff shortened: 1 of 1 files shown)


MartinHjelmare (approved):

matter_server/server/device_controller.py
```diff
@@ -587,26 +587,18 @@ async def remove_node(self, node_id: int) -> None:
             Clusters.OperationalCredentials.Attributes.CurrentFabricIndex,
         )
         fabric_index = node.attributes[attribute_path]
-
-        result: Clusters.OperationalCredentials.Commands.NOCResponse = (
+        try:
             await self.chip_controller.SendCommand(
```
agners, line 591: So this adds an exception handler. Before #500 we did not use an exception handler as well.

It probably make sense to have an exception handler in case the command fails.

But the command can also succeed, but return an error no?

agners: @marcelveldt ah I think this show up now because we log the error properly (since #499)! However, the exceptions have been thrown always, just now they are visible in the Server logs.

So the correct PR description here is "Handle ChipStackError on Node removal properly".

But I'd still prefer if you can check the result of the successful call as I've implemented in #500.

marcelveldt: Nope, it doesnt return the error code, that was the whole error

marcelveldt: For more clarity: #500 introduced a bug because the result of SendCommand is None, there is no NOCResponse returned. Instead it will raise an exception if the command fails

agners: > For more clarity: https://github.com/home-assistant-libs/python-matter-server/pull/500 introduced a bug because the result of SendCommand is None

I've tested that code, at least in the success case this was not true.

> Nope, it doesnt return the error code, that was the whole error

Can you please elaborate?

As per my testing, a successful RemoveFabric command returns a result with `NodeOperationalCertStatusEnum.kOk`. I expect two possible errors: The device is on the network and reachable, but the command fails. In this case I expect the result to be not `bOK`. The device is no longer reachable (typically because it got reset). In this case I expect an exception.

The code currently only handles the exception case. Why can't we re-add the first case?

marcelveldt: Looks like the result can be None or an int so I've adjusted it a bit again. Now it can handle both and have exception handling

agners: Hm, might be dependent on the version/build of the SDK :thinking: . LGTM.

PR #508, opened 2024-01-25
Add diagnostics for matter nodes

marcelveldt:
Add models and functions to get (and refresh) diagnostic details for Matter nodes:

- Add ping_node command to the server to refresh and ping all ip addresses of a node and return them in a human readable format.
- Adjust matter_fabrics helper in the client to refresh the data
- Add node_diagnostics helper in the client to return a collection of node diagnostics in a human readable format, already prepared in a custom format that is more ready suitable for HA (like string enums that can be used as translation keys)

### DIFF
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -318,7 +318,4 @@ def convert_ip_address(hex_ip: str | bytes, ipv6: bool = False) -> str:
     if isinstance(hex_ip, str):
         # note that the bytes string can be optionally base64 encoded
         hex_ip = base64.b64decode(hex_ip)
-    if ipv6:
-        hex_str = "".join(f"{byte:02x}" for byte in hex_ip)
-        return ":".join(hex_str[i : i + 4] for i in range(0, len(hex_str), 4))
-    return socket.inet_ntoa(hex_ip)
+    return socket.inet_ntop(socket.AF_INET6 if ipv6 else socket.AF_INET, hex_ip)
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -1,6 +1,7 @@
 """Utils for Matter server (and client)."""
 from __future__ import annotations
 
+import base64
 from base64 import b64decode
 import binascii
 from dataclasses import MISSING, asdict, fields, is_dataclass
@@ -10,6 +11,7 @@ from functools import cache
 from importlib.metadata import PackageNotFoundError, version as pkg_version
 import logging
 import platform
+import socket
 from types import NoneType, UnionType
 from typing import (
     TYPE_CHECKING,
@@ -219,8 +221,7 @@ def parse_value(
     ):
         return uint(value)
     if value_type is float32 and (
-        isinstance(value, float)
-        or isinstance(value, int)
+        isinstance(value, (float, int))
         or (isinstance(value, str) and value.isnumeric())
     ):
         return float32(value)
@@ -291,3 +292,33 @@ def chip_core_version() -> str:
         # TODO: Fix this once we can install our own wheels on macos.
         return chip_clusters_version()
     return package_version(CHIP_CORE_PKG_NAME)
+
+
+def convert_hex_string(hex_str: str | bytes) -> str:
+    """Convert hex string received from the sdk to a regular (unicode) string."""
+    if isinstance(hex_str, str):
+        # note that the bytes string can be optionally base64 encoded
+        # when we send it back and forth over our api
+        hex_str = base64.b64decode(hex_str)
+
+    return "".join(f"{byte:02x}" for byte in hex_str)
+
+
+def convert_mac_address(hex_mac: str | bytes) -> str:
+    """Convert hexadecimal MAC received from the sdk to a regular mac-address string."""
+    if isinstance(hex_mac, str):
+        # note that the bytes string can be optionally base64 encoded
+        hex_mac = base64.b64decode(hex_mac)
+
+    return ":".join("{:02x}".format(byte) for byte in hex_mac)  # pylint: disable=C0209
+
+
+def convert_ip_address(hex_ip: str | bytes, ipv6: bool = False) -> str:
+    """Convert hexadecimal IP received from the sdk to a regular IP string."""
+    if isinstance(hex_ip, str):
+        # note that the bytes string can be optionally base64 encoded
+        hex_ip = base64.b64decode(hex_ip)
+    if ipv6:
+        hex_str = "".join(f"{byte:02x}" for byte in hex_ip)
+        return ":".join(hex_str[i : i + 4] for i in range(0, len(hex_str), 4))
+    return socket.inet_ntoa(hex_ip)
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -295,7 +295,7 @@ def chip_core_version() -> str:
 
 
 def convert_hex_string(hex_str: str | bytes) -> str:
-    """Convert hex string received from the sdk to a regular (unicode) string."""
+    """Convert (Base64 encoded) byte array received from the sdk to a regular (unicode) string."""
     if isinstance(hex_str, str):
         # note that the bytes string can be optionally base64 encoded
         # when we send it back and forth over our api
@@ -305,7 +305,7 @@ def convert_hex_string(hex_str: str | bytes) -> str:
 
 
 def convert_mac_address(hex_mac: str | bytes) -> str:
-    """Convert hexadecimal MAC received from the sdk to a regular mac-address string."""
+    """Convert (Base64 encoded) byte array MAC received from the sdk to a regular mac-address."""
     if isinstance(hex_mac, str):
         # note that the bytes string can be optionally base64 encoded
         hex_mac = base64.b64decode(hex_mac)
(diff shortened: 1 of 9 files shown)


matter_server/common/models.py
```diff
@@ -106,6 +109,74 @@ class ServerDiagnostics:
     events: list[dict]
 
 
+class NodeType(Enum):
+    """Enum with Matter node types."""
```
agners, line 113: This is somewhat confusing. While this are values stored in Matter enums, these note types are a pure Thread concept.

marcelveldt: I'd rather not as those are thread specific only. I want to make one value regardless if its wifi, thread or bridge

matter_server/common/models.py
```diff
@@ -106,6 +109,74 @@ class ServerDiagnostics:
...
+    """Representation of a ActiveFabric message."""
+
+    fabric_index: int
+    vendor_id: int
+    fabric_id: int
+    name: str
+
+
+@dataclass
+class NodeDiagnostics:
```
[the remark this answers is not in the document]
marcelveldt: I don't agree as this is what makes it more useable/readable. A very small abstraction layer on top of those huge clusters. If you want the pure/raw data, just use read_attribute. This is all about creating a unified diagnostics object that we can easily use on the HA frontend.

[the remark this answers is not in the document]
marcelveldt: Note that we are already parsing some of the data into our own format (such as ip addresses)

matter_server/server/helpers/utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+async def ping_ip(ip_address: str) -> bool:
+    """Ping given (IPv4 or IPv6) IP-address."""
+    is_ipv6 = ":" in ip_address
+    if is_ipv6 and PLATFORM_MAC:
+        # macos does not have support for -W (timeout) on ping6 ?!
+        cmd = f"ping6 -c 1 {ip_address}"
+    elif is_ipv6:
+        cmd = f"ping6 -c 1 -W 2 {ip_address}"
```
marcelveldt, line 38: should we maybe remove the 2 second timeout here because theoretically, battery powered devices may take multiple seconds to respond ?

matter_server/server/device_controller.py
```diff
@@ -662,6 +677,171 @@ def resubscribe() -> None:
...
+    async def node_diagnostics(self, node_id: int) -> NodeDiagnostics:
+        """Return diagnostics for the given node."""
+        node = self.get_node(node_id)
+        # ping all interfaces (will also refresh data)
+        ping_result = await self.ping_node(node_id)
+        # grab some details from the first (operational) network interface
+        attribute = Clusters.GeneralDiagnostics.Attributes.NetworkInterfaces
+        attr_path = f"0/{attribute.cluster_id}/{attribute.attribute_id}"
+        attr_data = node.attributes.get(attr_path)
+        attr_data = cast(list[dict[str, Any]], node.attributes.get(attr_path))
```
MartinHjelmare, line 785: Duplicate line.

MartinHjelmare: Do we need to use `dict.get`?

tests/server/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    assert convert_mac_address("ji4yiD/r91c=") == "8e:2e:32:88:3f:eb:f7:57"
+
+    b"\xfe\x80\x00\x00\x00\x00\x00\x00\x04P\x1dB]@\x8eE"
+
+
+async def test_ping() -> None:
+    """Test ping ip util."""
+    # test valid ipv4
+    assert await ping_ip("127.0.0.1") is True
```
MartinHjelmare, line 25: Maybe patch `create_subprocess_shell`?

marcelveldt: yeah, I was in doubt about it as I really wanted to test the output here so perform the actual call and not patch it

matter_server/server/device_controller.py
marcelveldt: fixed

matter_server/server/helpers/utils.py
marcelveldt: to answer my own question - I have adjusted it. Use 10 second timeout for sleepy devices and 2 for all other

matter_server/common/models.py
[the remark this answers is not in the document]
agners: > I don't agree as this is what makes it more useable/readable. A very small abstraction layer on top of those huge clusters. If you want the pure/raw data, just use read_attribute. This is all about creating a unified diagnostics object that we can easily use on the HA frontend.

But if we start doing that here, why don't we do that for lights, vacuums and other things?

> So to give a more concrete answer: I am not conflating the WiFi and Thread diagnostics clusters. I'm just taking parts from it and model it into our own very small abstraction layer in a format useable for quick diagnostics overview of a node.

I agree, we need to simplify this for the frontend. And that is where the code belongs: Client side! We don't need to make our WS API more complex. We already do exchange data using  the Cluster objects, so why not stay consistent?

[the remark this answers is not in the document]
marcelveldt: I think you are overreacting here. I really do not get your issue here.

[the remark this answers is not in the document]
marcelveldt: As discussed privately; I'm are going to split this up into two parts. Ping will be server side in the custom modelled format. The diagnostics modeling will stay but client only so that the server can stay as simple/flat as possible

matter_server/common/models.py
marcelveldt: as discussed this will be moved to client side only (just for HA)

matter_server/common/helpers/util.py
```diff
@@ -291,3 +292,33 @@ def chip_core_version() -> str:
...
+    """Convert hexadecimal MAC received from the sdk to a regular mac-address string."""
+    if isinstance(hex_mac, str):
+        # note that the bytes string can be optionally base64 encoded
+        hex_mac = base64.b64decode(hex_mac)
+
+    return ":".join("{:02x}".format(byte) for byte in hex_mac)  # pylint: disable=C0209
+
+
+def convert_ip_address(hex_ip: str | bytes, ipv6: bool = False) -> str:
+    """Convert hexadecimal IP received from the sdk to a regular IP string."""
```
agners, line 317: Nit
```suggestion
    """Convert Base64 encoded byte array IP received from the Matter SDK to a regular IP string."""
```

matter_server/common/helpers/util.py
```diff
@@ -291,3 +292,33 @@ def chip_core_version() -> str:
...
+    if isinstance(hex_ip, str):
+        # note that the bytes string can be optionally base64 encoded
+        hex_ip = base64.b64decode(hex_ip)
+    if ipv6:
+        hex_str = "".join(f"{byte:02x}" for byte in hex_ip)
+        return ":".join(hex_str[i : i + 4] for i in range(0, len(hex_str), 4))
+    return socket.inet_ntoa(hex_ip)
```
agners, line 324:
```suggestion
    return socket.inet_ntop(socket.AF_INET6 if ipv6 else socket.AF_INET, hex_ip)
```

matter_server/server/device_controller.py
```diff
@@ -662,6 +670,69 @@ def resubscribe() -> None:
...
+        if node.available:
+            # try to refresh the GeneralDiagnostics.NetworkInterface attribute
+            # so we have the most accurate information before pinging
+            try:
+                await self.read_attribute(node_id, attr_path)
+            except (NodeNotResolving, ChipStackError) as err:
+                LOGGER.exception(err)
+
+        battery_powered = (
+            node.attributes.get("0/53/1", 0)
```
agners, line 694: I guess the idea is to get the cached value here? It is kinda annoying that we have to use the numerical representation here :thinking:

agners (changes requested):

matter_server/common/helpers/util.py
marcelveldt: well, its only base64 encoded if we've sent it over the wire (as we do turn bytes into base64 strings) so the docstring was a bit confusing :-)

matter_server/server/device_controller.py
marcelveldt: yeah, this is a fast key lookup, otherwise I need to look it up from the class+attribute

marcelveldt: I'll put it in a constant

matter_server/common/helpers/util.py
agners: From what I can tell, also the input of this function is base64, e.g. from the test you call it with:

```
convert_ip_address("KgARtxIxhABW70T//kmvxg==", True)
                    ^ this is the IP address represented as byte array encoded as a base64 string.
```

matter_server/server/device_controller.py
marcelveldt: created a constant

matter_server/common/helpers/util.py
marcelveldt: it accepts both - if its a string, its base64 encoded

marcelveldt: Note that we convert it to a base64 string when sending it over the API as we cant send byte sin json

agners: Right :+1:  LGTM!

agners (approved):
LGTM

PR #515, opened 2024-01-29
Guard node resolving with a lock

marcelveldt:
It's still not entirely clear but it looks like multiple resolves at the same time is not very safe and one action may even be discarded if another attempt comes in. So for example if the server is busy trying to resolve an offline node it may cause commissioning to fail. So for safety reasons put a lock on the resolving. It makes init of nodes somewhat slower but its safer.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -107,7 +107,7 @@ class MatterDeviceController:
         self.thread_credentials_set: bool = False
         self.compressed_fabric_id: int | None = None
         self._node_lock: dict[int, asyncio.Lock] = {}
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -107,6 +107,7 @@ def __init__(
         self.thread_credentials_set: bool = False
         self.compressed_fabric_id: int | None = None
         self._node_lock: dict[int, asyncio.Lock] = {}
+        self._resolve_lock: asyncio.Lock = asyncio.Lock()
```
MartinHjelmare, line 110:
```suggestion
        self._resolve_lock = asyncio.Lock()
```

MartinHjelmare (approved):

PR #516, opened 2024-01-30
Fix model for CommissioningParameters in client

marcelveldt:
The server returns a CommissioningParameters object on the "open_commissioning_window" which is a dataclass specified in the chip devicecontroller.

The typing on our python client side was still a tuple, so wrong.

Copied the CommissioningParameters dataclass to our shared models for the client.

### DIFF
--- a/matter_server/common/models.py
+++ b/matter_server/common/models.py
@@ -172,3 +172,16 @@ MessageType = (
     | ErrorResultMessage
     | ServerInfoMessage
 )
+
+
+@dataclass
+class CommissioningParameters:
+    """
+    Object that is returned on the 'open_commisisoning_window' command.
+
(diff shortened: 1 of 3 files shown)


frenck (approved):

matter_server/common/models.py
```diff
@@ -172,3 +172,16 @@ class ServerInfoMessage:
...
+
+@dataclass
+class CommissioningParameters:
+    """
+    Object that is returned on the 'open_commisisoning_window' command.
+
+    NOTE: This is just a copy of the dataclass specified in chip.ChipDeviceCtrl
+    """
+
+    setupPinCode: int  # pylint: disable=invalid-name
```
MartinHjelmare, line 185: We don't use camelCase attribute names for any of our other models.

PR #518, opened 2024-01-30
Speedup diagnostics

marcelveldt:
After playing a bit with the diagnostics fetching while implementing it in the frontend, I quickly came to the conclusion that it was not working as intended, especially the retrieval of the values and ping slows it down badly, especially for thread devices. So some changes;

- Add IP addresses and fabrics to watched attributes (so we do not have to fetch them)
- Remove ping from diagnostics 
- Use available bool on diagnostics model
- no duplicate conversion needed for network name

### DIFF
--- a/matter_server/server/helpers/utils.py
+++ b/matter_server/server/helpers/utils.py
@@ -3,6 +3,8 @@
 import asyncio
 import platform
 
+import async_timeout
+
 PLATFORM_MAC = platform.system() == "Darwin"
 
 
@@ -16,7 +18,13 @@ async def ping_ip(ip_address: str, timeout: int = 2) -> bool:
         cmd = f"ping6 -c 1 -W {timeout} {ip_address}"
     else:
         cmd = f"ping -c 1 -W {timeout} {ip_address}"
-    return (await check_output(cmd))[0] == 0
+    try:
+        # we add an additional timeout here as safeguard and to account for the fact
+        # that macos does not seem to have timeout on ping6
+        async with async_timeout.timeout(timeout + 2):
+            return (await check_output(cmd))[0] == 0
+    except (asyncio.TimeoutError, asyncio.CancelledError):
+        return False
 
 
 async def check_output(shell_cmd: str) -> tuple[int | None, bytes]:
--- a/matter_server/server/helpers/utils.py
+++ b/matter_server/server/helpers/utils.py
@@ -23,7 +23,7 @@ async def ping_ip(ip_address: str, timeout: int = 2) -> bool:
         # that macos does not seem to have timeout on ping6
         async with async_timeout.timeout(timeout + 2):
             return (await check_output(cmd))[0] == 0
-    except (asyncio.TimeoutError, asyncio.CancelledError):
+    except asyncio.TimeoutError:
         return False
 
 
@@ -34,5 +34,10 @@ async def check_output(shell_cmd: str) -> tuple[int | None, bytes]:
         stderr=asyncio.subprocess.STDOUT,
         stdout=asyncio.subprocess.PIPE,
     )
-    stdout, _ = await proc.communicate()
+    try:
+        stdout, _ = await proc.communicate()
+    except asyncio.CancelledError:
+        proc.terminate()
+        await proc.communicate()
+        raise
     return (proc.returncode, stdout)
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -260,9 +260,6 @@ class MatterClient:
             # ignore invalid/non-operational interfaces
             if not network_interface.isOperational:
                 continue
-            # ignore invalid/non-operational interfaces
-            if not network_interface.isOperational:
-                continue
             if (
                 network_interface.type
                 == Clusters.GeneralDiagnostics.Enums.InterfaceTypeEnum.kThread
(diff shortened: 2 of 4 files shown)


matter_server/server/helpers/utils.py
```diff
@@ -16,7 +18,13 @@ async def ping_ip(ip_address: str, timeout: int = 2) -> bool:
         cmd = f"ping6 -c 1 -W {timeout} {ip_address}"
     else:
         cmd = f"ping -c 1 -W {timeout} {ip_address}"
-    return (await check_output(cmd))[0] == 0
+    try:
+        # we add an additional timeout here as safeguard and to account for the fact
+        # that macos does not seem to have timeout on ping6
+        async with async_timeout.timeout(timeout + 2):
+            return (await check_output(cmd))[0] == 0
+    except (asyncio.TimeoutError, asyncio.CancelledError):
```
MartinHjelmare, line 26: What raises CancelledError?

marcelveldt: So, 'proc.communicate' raises CancelledError when aborted with the timeout

MartinHjelmare: `asyncio.CancelledError` is send into the timeout context manager and `asyncio.TimeoutError` is raised outside the context manager. If there's a CancelledError raised outside the timeout context manager I think it means that the task was cancelled from outside.

```py
Python 3.11.7 (main, Dec  8 2023, 18:56:57) [GCC 9.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import asyncio

In [2]: async def main():
   ...:     try:
   ...:         async with asyncio.timeout(2):
   ...:             try:
   ...:                 await asyncio.sleep(3)
   ...:             except asyncio.CancelledError:
   ...:                 print("got cancelled!")
   ...:                 raise
   ...:     except asyncio.TimeoutError:
   ...:         print("timeout!")
   ...:     else:
   ...:         print("finish ok!")
   ...: 

In [3]: asyncio.run(main())
got cancelled!
timeout!

In [4]: async def cancel_from_outside(coro):
   ...:     task = asyncio.create_task(coro)
   ...:     await asyncio.sleep(0)
   ...:     task.cancel()
   ...:     try:
   ...:         await task
   ...:     except asyncio.CancelledError:
   ...:         print("task was cancelled and raised CancelledError!")
   ...: 
   ...: 

In [5]: asyncio.run(cancel_from_outside(main()))
got cancelled!
task was cancelled and raised CancelledError!

In [6]: async def cancel_from_outside(coro):
   ...:     task = asyncio.create_task(coro)
   ...:     await asyncio.sleep(4)
   ...:     task.cancel()
   ...:     try:
   ...:         await task
   ...:     except asyncio.CancelledError:
   ...:         print("task was cancelled and raised CancelledError!")
   ...: 

In [7]: asyncio.run(cancel_from_outside(main()))
got cancelled!
timeout!

In [8]:
```

marcelveldt: I run into it very consistently, hence I added the cancellederror here. Maybe its a simple race condition ?

MartinHjelmare: If I test a simplified ping script I get a warning if the timeout context manager times out before the subprocess times out.

```py
In [14]: asyncio.run(ping_ip("192.168.0.132"))
Out[14]: False

Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 99812 is closed
```

Reading the asyncio.subprocess docs. Maybe we should catch `asyncio.CancelledError` around `process.communicate`, terminate the process, wait for the process to terminate and re-raise `asyncio.CancelledError`?

```py
Python 3.11.7 (main, Dec  8 2023, 18:56:57) [GCC 9.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import asyncio

In [2]: async def ping_ip(ip_address: str, timeout: int = 10) -> bool:
   ...:     cmd = f"ping -c 1 -W {timeout} {ip_address}"
   ...:     try:
   ...:         async with asyncio.timeout(timeout - 9):
   ...:             return (await check_output(cmd))[0] == 0
   ...:     except asyncio.TimeoutError:
   ...:         print("check output timed out")
   ...:         return False
   ...: 

In [3]: async def check_output(shell_cmd: str) -> tuple[int | None, bytes]:
   ...:     proc = await asyncio.create_subprocess_shell(
   ...:         shell_cmd,
   ...:         stderr=asyncio.subprocess.STDOUT,
   ...:         stdout=asyncio.subprocess.PIPE,
   ...:     )
   ...:     try:
   ...:         stdout, _ = await proc.communicate()
   ...:     except asyncio.CancelledError:
   ...:         print("communicate was cancelled, terminate process")
   ...:         proc.terminate()
   ...:         await proc.communicate()
   ...:         raise
   ...:     return (proc.returncode, stdout)
   ...: 

In [4]: asyncio.run(ping_ip("192.168.0.132"))
communicate was cancelled, terminate process
check output timed out
Out[4]: False

In [5]: 
```

MartinHjelmare: In production the event loop won't be closed probably, but it's an indication that we haven't cleaned up the task before returning, I think.

marcelveldt: applied your suggestion and that indeed fixes it, thanks @MartinHjelmare !

MartinHjelmare (approved):
Looks good!

PR #520, opened 2024-01-30
Use our own global models for CommissionableNodeData and Commissionin…

marcelveldt:
Use our own global models for CommissionableNodeData and CommissioningParameters instead of the (non camelcase) ones from the SDK

These were not yet used so this is not a breaking change but I've bumped the schema anyways.


MartinHjelmare (approved):
Thanks!

PR #526, opened 2024-02-05
Support specifying which IP addresses to listen on

agners:
Allow to define which IP addresses to bind the WebSocket server on. This is especially useful to limit which addresses the add-on will listen on since it runs in host network context.

### DIFF
--- a/matter_server/server/__main__.py
+++ b/matter_server/server/__main__.py
@@ -52,7 +52,7 @@ parser.add_argument(
     type=str,
     action="append",
     default=DEFAULT_LISTEN_ADDRESS,
-    help=f"IP address to bind the websocket server to, defaults to listen on any interface (IPv4 and IPv6).",
+    help="IP address to bind the websocket server to, defaults to listen on any interface (IPv4 and IPv6).",
 )
 parser.add_argument(
     "--log-level",
--- a/matter_server/server/__main__.py
+++ b/matter_server/server/__main__.py
@@ -13,6 +13,8 @@ from .server import MatterServer
 DEFAULT_VENDOR_ID = 0xFFF1
 DEFAULT_FABRIC_ID = 1
 DEFAULT_PORT = 5580
+# Default to None to bind to all addresses on both IPv4 and IPv6
+DEFAULT_LISTEN_ADDRESS = None
 DEFAULT_STORAGE_PATH = os.path.join(Path.home(), ".matter_server")
 
 # Get parsed passed in arguments.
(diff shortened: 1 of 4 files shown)


matter_server/server/server.py
```diff
@@ -62,13 +62,15 @@ def __init__(
         vendor_id: int,
         fabric_id: int,
         port: int,
+        listen_addresses: list[str],
```
MartinHjelmare, line 65:
```suggestion
        listen_addresses: list[str] | None = None,
```

matter_server/server/server.py
```diff
@@ -62,13 +62,15 @@ def __init__(
...
         primary_interface: str | None,
```
MartinHjelmare, line 66:
```suggestion
        primary_interface: str | None = None,
```

matter_server/server/server.py
```diff
@@ -62,13 +62,15 @@ def __init__(
...
-        primary_interface: str | None,
+        listen_addresses: list[str] | None = None,
+        primary_interface: str | None = None,
     ) -> None:
         """Initialize the Matter Server."""
         self.storage_path = storage_path
         self.vendor_id = vendor_id
         self.fabric_id = fabric_id
-        self.port = port
+        self._port = port
```
MartinHjelmare, line 72: Was this change intended? Technically a breaking change but the attribute is probably not needed for outside use, besides for debugging.

agners: My thought was to align it with `listen_address`, which I preferred to be private. I could a read only property instead? :thinking:

marcelveldt: I'd leave it public

matter_server/server/server.py
```diff
@@ -62,13 +62,15 @@ def __init__(
         vendor_id: int,
         fabric_id: int,
         port: int,
-        primary_interface: str | None,
+        listen_addresses: list[str] | None = None,
```
marcelveldt, line 65: is it indended to have a list here ?

matter_server/server/__main__.py
```diff
@@ -95,6 +104,7 @@ def main() -> None:
         int(args.vendorid),
         int(args.fabricid),
         int(args.port),
+        args.listen_address,
```
marcelveldt, line 107: you are sending a None or single string value here while the typing of "listen_addresses" is set to a list of strings or None.

matter_server/server/server.py
agners: Yes, you may listen on multiple addresses or address families. E.g. to explicitly listen to what is the default today you'd pass `--listen-address 0.0.0.0 --listen-address ::`.

matter_server/server/__main__.py
marcelveldt: nvm, I overlooked the append.

matter_server/server/server.py
marcelveldt: Yeah, see above, I overlooked the append. All good.

marcelveldt (approved):

marcelveldt:
see the linter error from mypy

agners:
> see the linter error from mypy

Ugh, yeah I see `TCPSite` supports only a single string as host, whereas the underlying `loop.create_server` supports a list. We've run into this limitation in the Core already (see https://github.com/home-assistant/core/blob/2024.2.0b8/homeassistant/components/http/web_runner.py#L14).

I've now added a similar implementation here too.

@bdraco do you think we can add something like that to aiohttp? :pleading_face:

marcelveldt (approved):

Issue #523, opened 2024-02-01
Received non-Text message


olavt:
From time to time, I see these kind of messages in the log:

2024-02-01 13:59:44 matter-server matter_server.server.client_handler[1] WARNING [548415031568] Disconnected: Received non-Text message.

As far as I know, I never send non-text messages to the server.

What could be causing these?

Maybe a good idea to log some of the content received?

PR #530, opened 2024-02-06
Log non-text messages in websocket api

marcelveldt:
Log non-text messages in websocket api

Addresses #523

### DIFF
--- a/matter_server/server/client_handler.py
+++ b/matter_server/server/client_handler.py
@@ -100,7 +100,7 @@ class WebsocketClientHandler:
                     break
 
                 if msg.type != WSMsgType.TEXT:
-                    self._logger.debug("Received non-Text message: %s", msg.data)
+                    self._logger.warning("Received non-Text message: %s", msg.data)
                     continue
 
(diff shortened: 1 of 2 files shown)


matter_server/server/client_handler.py
```diff
@@ -91,13 +92,17 @@ async def handle_client(self) -> web.WebSocketResponse:
...
                     break
 
-                if msg.type != WSMsgType.TEXT:
-                    disconnect_warn = "Received non-Text message."
+                if msg.type == WSMsgType.ERROR:
+                    disconnect_warn = f"Received error message: {msg.data}"
                     break
 
+                if msg.type != WSMsgType.TEXT:
+                    self._logger.debug("Received non-Text message: %s", msg.data)
```
agners, line 103: Probably the case at hand was a `WSMsgType.CLOSED` message. Since we only support text messages, this really is unexpected, so I'd go for warning:

```suggestion
                    self._logger.warning("Received non-Text message: %s", msg.data)
```

agners (approved):

PR #531, opened 2024-02-07
Implement zeroconf for operational node discovery

marcelveldt:
Implemented zeroconf to do our own mdns discovery of operational (and later commissionable) Matter nodes.
This solves a number of issues, most importantly that we do no longer have to rely on the (bad) polling behavior of the SDK if a node goes unavailable. Instead we can detect it coming back online through mdns discovery proactively.

- Initialize nodes on discovery over mdns
- Do no longer try to resolve all nodes at startup (as only the live ones will be discovered over mdns now).
- Mark node offline if zeroconf reports it as offline.
- Prepare listening for commissionable nodes, to be picked up in a follow-up PR.
- Some small polishes such as deny commands to an unavailable node

Fixes:

- Continuous error logging of the SDK trying to resolve offline nodes.
- If a Matter device is removed from power it took hours before it was rediscovered when power was restored, this is now seconds.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -958,10 +958,9 @@ class MatterDeviceController:
             # we debounce it a bit so we only mark the node unavailable
             # at the second resubscription attempt
             if node.available and self._last_subscription_attempt[node_id] >= 1:
-                node.available = False
                 # NOTE: if the node is (re)discovered by mdns, that callback will
                 # take care of resubscribing to the node
-                self.server.signal_event(EventType.NODE_UPDATED, node)
+                asyncio.create_task(self._node_offline(node_id))
             self._last_subscription_attempt[node_id] += 1
 
         def resubscription_succeeded(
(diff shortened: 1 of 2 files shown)


MartinHjelmare (approved):

agners:
> on the (bad) polling behavior of the SDK if a node goes unavailable

What polling behavior are you referring to?

The rescheduled interview when a node stopped being available (not resolving) was on our end no?

matter_server/server/device_controller.py
```diff
@@ -1166,3 +1247,19 @@ def _write_node_state(self, node_id: int, force: bool = False) -> None:
...
         )
+
+    async def _node_offline(self, node_id: int) -> None:
+        """Mark node as offline."""
+        # Remove and cancel any existing interview/subscription reschedule timer
+        if existing := self._sub_retry_timer.pop(node_id, None):
+            existing.cancel()
+        # shutdown existing subscriptions
+        if sub := self._subscriptions.pop(node_id, None):
+            await self._call_sdk(sub.Shutdown)
```
agners, line 1258: This code is also used when deleting the node, I think we should extract it in a common function.

matter_server/server/device_controller.py
```diff
@@ -937,8 +958,9 @@ def resubscription_attempted(
             # we debounce it a bit so we only mark the node unavailable
             # at the second resubscription attempt
             if node.available and self._last_subscription_attempt[node_id] >= 1:
-                node.available = False
-                self.server.signal_event(EventType.NODE_UPDATED, node)
+                # NOTE: if the node is (re)discovered by mdns, that callback will
+                # take care of resubscribing to the node
+                asyncio.create_task(self._node_offline(node_id))
```
agners, line 963: I am a bit worried that we get false positives here. But let's see.

Btw, you should store the task somewhere, see also https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task and https://bugs.python.org/issue44665.

> **Important:** Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

PR #535, opened 2024-02-07
Fix networkname retrieval for some wifi devices

marcelveldt:


matter_server/client/client.py
```diff
@@ -338,8 +341,25 @@ async def node_diagnostics(self, node_id: int) -> NodeDiagnostics:
...
+        if not network_name and (
+            networks := node.get_attribute_value(
+                0,
+                cluster=None,
+                attribute=Clusters.NetworkCommissioning.Attributes.Networks,
+            )
+        ):
+            for network in networks:
+                if isinstance(network.networkID, bytes):
+                    network_name = network.networkID.decode("utf-8")
```
agners, line 358: So we just pick the first network from the `NetworkCommissioning` cluster? That could be any WiFi network no? I think this could lead to more problems/confusion then being helpful.

marcelveldt: nope, it is a network that has been used to commisission the device. So 99% sure this network is active on the device.

FYI: From the devices I have tested;

- some only had the network in the lastNetworkID
- Some only had the network in this list

But it was never the wrong network

marcelveldt: It is not ANY network. There is another clustercommand to let the device do an active scan for networks and return those

agners: Ah I see those are possible options. I see there is `Connected` property, we could use that to make the best possible guess?

marcelveldt: good suggestion, I'll add that

marcelveldt: done

agners (approved):

PR #536, opened 2024-02-07
Treat mdns update events as add events as safeguard

marcelveldt:
Just as a safe guard, treat mdns update events as add events just to catch the edge case where we somehow missed the add message.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1183,9 +1183,6 @@ class MatterDeviceController:
         name: str,
         state_change: ServiceStateChange,
     ) -> None:
-        if state_change not in (ServiceStateChange.Added, ServiceStateChange.Removed):
-            # we're not interested in update messages so return early
-            return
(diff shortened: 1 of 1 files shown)


agners (approved):

matter_server/server/device_controller.py
```diff
@@ -1211,7 +1208,7 @@ async def _on_mdns_operational_node_state(
         try:
             self._mdns_inprogress.add(node_id)
             node = self._nodes[node_id]
-            if state_change == ServiceStateChange.Added:
+            if state_change in (ServiceStateChange.Added, ServiceStateChange.Updated):
```
MartinHjelmare, line 1211: Do we need the same change on line 1229?

marcelveldt: Let's look at that once we do the follow up PR about commissionable node discovery.

MartinHjelmare (approved):

PR #537, opened 2024-02-07
Process the mdns messages semi parallel

marcelveldt:
Adjust the processing of mdns events for operational matter nodes a small bit.
We want to process them in sequential order but at the same time not block processing if one node is just slow to respond or whatever.

This introduces a queue and queue processing task per node (or strictly speaking, mdns name)

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -137,10 +137,10 @@ class MatterDeviceController:
         self.chip_controller = self.server.stack.fabric_admin.NewController(
             paaTrustStorePath=str(PAA_ROOT_CERTS_DIR)
         )
-        self.compressed_fabric_id = await self._call_sdk(
-            self.chip_controller.GetCompressedFabricId
+        self.compressed_fabric_id = cast(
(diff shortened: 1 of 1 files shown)


agners (approved):
The lifetime of the asyncio tasks is better managed, I like this approach :+1:

marcelveldt:
Adjusted the code a bit to fix the (strangely) failing pytest

agners (approved):

matter_server/server/device_controller.py
```diff
@@ -1189,38 +1196,46 @@ def _on_mdns_service_state_change(
...
-    async def _on_mdns_operational_node_state(
-        self, name: str, state_change: ServiceStateChange
+            name = name.lower()
+            if not name.startswith(self.fabric_id_hex):
+                # filter out messages that are not for our fabric
+                return
+            if existing := self._mdns_queues.get(name):
+                queue = existing[0]
+            else:
+                # we want mdns messages to be processes sequentially PER NODE but in
```
MartinHjelmare, line 1206: processed

PR #538, opened 2024-02-07
Refactor the interview logic a small bit

marcelveldt:
A full re-interview was required when the API schema was bumped, which was not correct.
We only need to force a re-interview when the node data changed, for example when we bumped the sdk or changed the datamodel.

Also a bit of polishing of the interview logic itself.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1051,8 +1051,7 @@ class MatterDeviceController:
         )
 
     async def _setup_node(self, node_id: int) -> None:
-        """Setup subscriptions and interview (if needed) for known/discovered node."""
-
+        """Handle set-up of subscriptions and interview (if needed) for known/discovered node."""
         if node_id not in self._nodes:
             raise NodeNotExists(f"Node {node_id} does not exist.")
 
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -47,7 +47,7 @@ from ..common.models import (
     MatterNodeEvent,
     NodePingResult,
 )
-from .const import PAA_ROOT_CERTS_DIR
+from .const import DATA_MODEL_SCHEMA_VERSION, PAA_ROOT_CERTS_DIR
 from .helpers.paa_certificates import fetch_certificates
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


matter_server/server/device_controller.py
```diff
@@ -1050,48 +1050,30 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
+        node_data = self._nodes[node_id]
         if (
-            node_data is None
+            # re-interview if we dont have any node attributes (empty node)
+            not node_data.attributes
             # re-interview if the schema has changed
-            or node_data.interview_version < SCHEMA_VERSION
+            or node_data.interview_version < DATA_MODEL_SCHEMA_VERSION
+            # re-interview if the last interview was too long ago
+            or (datetime.utcnow() - node_data.last_interview) > timedelta(days=90)
```
agners, line 1066: Any specific idea behind this?

Maybe (another/better) indicator would be software version? :thinking:

marcelveldt: yeah I dunno, it somehow felt good but at the same time meh... I'll remove it.
software version is tracked elsewhere

matter_server/server/device_controller.py
```diff
@@ -1050,48 +1050,30 @@ async def _call_sdk(self, func: Callable[..., _T], *args: Any, **kwargs: Any) ->
...
         # (re)interview node (only) if needed
-        node_data = self._nodes.get(node_id)
+        node_data = self._nodes[node_id]
         if (
-            node_data is None
+            # re-interview if we dont have any node attributes (empty node)
+            not node_data.attributes
             # re-interview if the schema has changed
-            or node_data.interview_version < SCHEMA_VERSION
+            or node_data.interview_version < DATA_MODEL_SCHEMA_VERSION
```
agners, line 1064: People which installed the beta have now interview version set to 7. This means next time we bump `DATA_MODEL_SCHEMA_VERSION` they won't reinterview, which could be problematic :grimacing: 

Two ways: Next time we need to jump to 8, or we use != here... I think I prefer != here, so we can also downgrade. Thoughts?

matter_server/server/device_controller.py

matter_server/server/device_controller.py
marcelveldt: lets go for the non equals

PR #540, opened 2024-02-08
Use previous available state

marcelveldt:


matter_server/server/device_controller.py
```diff
@@ -1244,8 +1244,6 @@ async def _node_offline(self, node_id: int) -> None:
             await self._call_sdk(sub.Shutdown)
         # mark node as unavailable
         node = self._nodes[node_id]
-        if not node.available:
-            return  # nothing to do to
```
agners, line 1248: Since available is the only change we do this node, I still feel we should only signal an update if this is indeed changing?

I guess we don't expect for `_node_offline` to get called twice typically (anymore), but it also shouldn't hurt? :thinking:

agners (approved):

TonyZZX:
Hi @marcelveldt could you revert this change?

Previous, many matter devices in my house would periodically reboot themselves, so they will be like unavailable for 10 seconds and then become available.
However, after updating to 5.2.2 (which contains this change), such devices would become unavailable all the times after they reboot and never be available, unless I manually reboot the matter addon.

marcelveldt:
No I would not. So you have flaky devices and then we need to revert this?

Did you already update to the latest version? A lot happened since this change

TonyZZX:
> No I would not. So you have flaky devices and then we need to revert this?
> 
> Did you already update to the latest version? A lot happened since this change

NVM, I don't think this is the root cause. This issue was introduced in 5.1.0, not 5.1.1

More discussion
https://github.com/home-assistant/addons/issues/3453

PR #543, opened 2024-02-08
Adjust the linter config

marcelveldt:
Revisit the linter config to some new sane defaults for a 2024 python project.

If we do settle on this config, we need to work through a small list of issues that is found by this new config


marcelveldt:
@MartinHjelmare what do you think ? Ok to switch to this new config ?
I'd like to use this confog for other python projects as well we maintain.

It will mean we'll have some work to fix the files according the issues found

pyproject.toml
```diff
@@ -39,106 +39,126 @@ server = [
...
+  "ruff==0.2.1",
+  "safety==3.0.1"
 ]
 
 [project.scripts]
 matter-server = "matter_server.server.__main__:main"
 
-[tool.codespell]
-ignore-words-list = "pase,"
+[tool.setuptools]
```
MartinHjelmare, line 59: I think we should sort this whole file :abcd: to make it easier to find things. My vscode sorts the the subitems automatically when I save a pyproject.toml file. It doesn't sort the top level items though. But I'd like for us to sort them manually if we go ahead here.

pyproject.toml
```diff
@@ -39,106 +39,126 @@ server = [
...
-mypy_path = "matter_server/"
 no_implicit_optional = true
-show_error_codes = true
+no_implicit_reexport = true
+strict_optional = true
 warn_incomplete_stub = true
+warn_no_return = true
 warn_redundant_casts = true
 warn_return_any = true
-warn_unreachable = true
```
MartinHjelmare, line 71: Why is this removed?

pyproject.toml
```diff
@@ -39,106 +39,126 @@ server = [
...
-    "ex",
-    "fp",
-    "i",
-    "id",
-    "j",
-    "k",
-    "Run",
-    "ip",
+  "_",
+  "ex",
```
MartinHjelmare, line 109: I don't think any of these are good names, besides `_` and possibly `T` as there some convention around those.

MartinHjelmare (commented):
I like this.

pyproject.toml
marcelveldt: do you know the name of the plugin you use for that ?

pyproject.toml
marcelveldt: accident I guess

marcelveldt:
OK, I will work more on this next week and start to lint the existing code. If we're happy with this config we can replicate it to the other projects in the libs org

pyproject.toml
MartinHjelmare: This one:
https://marketplace.visualstudio.com/items?itemName=be5invis.toml

marcelveldt:
This one is ready for a final check. Added a bunch of exceptions for the rules. We can resolve these one-by-one in follow up PR's.

pyproject.toml
```diff
@@ -28,117 +22,243 @@ dependencies = [
...
+  "codespell==2.2.6",
   "flake8==7.0.0",
   "flake8-docstrings==1.7.0",
   "isort==5.13.2",
   "mypy==1.8.0",
+  "pre-commit==3.6.0",
+  "pre-commit-hooks==4.5.0",
   "pylint==3.0.3",
   "pytest==7.4.4",
+  "pytest-asyncio==0.23.4",
```
MartinHjelmare, line 52:
```suggestion
  "pytest-asyncio==0.23.5",
```

pyproject.toml
```diff
@@ -28,117 +22,243 @@ dependencies = [
...
 warn_return_any = true
 warn_unreachable = true
 warn_unused_configs = true
 warn_unused_ignores = true
 
-[[tool.mypy.overrides]]
-ignore_missing_imports = true
-module = [
-  "aiorun",
+[tool.pylint.MASTER]
```
MartinHjelmare, line 101:
```suggestion
[tool.pylint.MAIN]
```

pyproject.toml
```diff
@@ -28,117 +22,243 @@ dependencies = [
...
+extension-pkg-whitelist = [
```
MartinHjelmare, line 102:
```suggestion
extension-pkg-allow-list = [
```

MartinHjelmare (changes requested):

MartinHjelmare (approved):
Nice!

PR #545, opened 2024-02-09
Fetch all Git certs and prefer them

priyankub:
Manufacturers release test certificates through the SDK (Git) as a part of their standard product release workflow. This will ensure those certs are correctly captured.

### DIFF
--- a/matter_server/server/helpers/paa_certificates.py
+++ b/matter_server/server/helpers/paa_certificates.py
@@ -16,7 +16,6 @@ from typing import List
 from aiohttp import ClientError, ClientSession
 from cryptography import x509
 from cryptography.hazmat.primitives import serialization
-import requests
 
 from matter_server.server.const import PAA_ROOT_CERTS_DIR
 
@@ -31,7 +30,7 @@ REPO = "connectedhomeip"
 PATH = "credentials/development/paa-root-certs"
 
 
-def get_directory_contents(owner: str, repo: str, path: str) -> List[str]:
+async def get_directory_contents(owner: str, repo: str, path: str) -> List[str]:
     """
     Fetch directory contents from a GitHub repository.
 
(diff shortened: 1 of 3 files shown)


matter_server/server/helpers/paa_certificates.py
```diff
@@ -11,21 +11,60 @@
 import logging
 from os import makedirs
 import re
+from typing import List
 
 from aiohttp import ClientError, ClientSession
 from cryptography import x509
 from cryptography.hazmat.primitives import serialization
+import requests
```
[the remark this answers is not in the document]
priyankub: Ack. will patch and share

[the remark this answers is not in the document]
priyankub: Switched to using aiohttp

matter_server/server/helpers/paa_certificates.py
```diff
@@ -22,10 +23,57 @@
...
+    Returns:
+        List[str]: A list of unique file names.
+    """
+    file_list = await get_directory_contents(OWNER, REPO, PATH)
+    # Filter out extension and remove duplicates
+    unique_file_names = list({file.split(".")[0] for file in file_list})
+    return unique_file_names
+
+
+GIT_CERTS = asyncio.run(get_git_file_list())
```
MartinHjelmare, line 74: I'd prefer fetching these as part of `fetch_git_certificates` so we don't have side effects on import of this module.

matter_server/server/helpers/paa_certificates.py
```diff
@@ -22,10 +23,57 @@
 PRODUCTION_URL = "https://on.dcl.csa-iot.org"
 TEST_URL = "https://on.test-net.dcl.csa-iot.org"
 GIT_URL = "https://github.com/project-chip/connectedhomeip/raw/master/credentials/development/paa-root-certs"  # pylint: disable=line-too-long
```
MartinHjelmare, line 25: We can use the git repo details below to construct this url.

matter_server/server/helpers/paa_certificates.py
priyankub: Ack. Made changes - testing it. Will deploy to patch after testing in my HA

matter_server/server/helpers/paa_certificates.py
priyankub: Ack. Testing changes now

matter_server/server/helpers/paa_certificates.py
```diff
@@ -11,24 +11,53 @@
...
+    Args:
+        owner (str): The owner of the GitHub repository.
+        repo (str): The name of the GitHub repository.
+        path (str): The path within the repository.
+
+    Returns:
+        list: A list of file names in the specified directory.
+    """
+    api_url = f"https://api.github.com/repos/{owner}/{repo}/contents/{path}"
+    async with ClientSession() as session:
```
MartinHjelmare, line 49: I suggest we inline this inside `fetch_git_certificates` since we already use a client session there. We can use the same session.

priyankub: Ack. Made the requested change

matter_server/server/helpers/paa_certificates.py
```diff
@@ -151,11 +125,23 @@ async def fetch_git_certificates() -> int:
...
-    file_list = await get_directory_contents(OWNER, REPO, PATH)
-    # Filter out extension and remove duplicates
-    git_certs = list({file.split(".")[0] for file in file_list})
+
     try:
         async with ClientSession(raise_for_status=True) as http_session:
+            # Fetch directory contents and filter out extension
+            api_url = f"https://api.github.com/repos/{OWNER}/{REPO}/contents/{PATH}"
+            async with http_session.get(api_url, timeout=20) as response:
+                if response.status == 200:
```
MartinHjelmare, line 134: We raise for status 400+. Are there other status codes that we expect that are errors?

matter_server/server/helpers/paa_certificates.py
```diff
@@ -151,11 +125,23 @@ async def fetch_git_certificates() -> int:
...
+                    contents = await response.json()
+                    git_certs = list({item["name"].split(".")[0] for item in contents})
```
MartinHjelmare, line 136: This could be kept as a set.

matter_server/server/helpers/paa_certificates.py
priyankub: Not that I know of. I intentionally kept redundant error handling. Do you suggest I remove `if response.status == 200` as we already check `raise_for_status=True` ?

MartinHjelmare: Yes, possibly.

MartinHjelmare (approved):
Thanks!

PR #547, opened 2024-02-14
Add an initial basic dashboard

balloob:
Add a basic dashboard for development. Based on [the work](https://github.com/wozz/matter-server-ui/tree/main) by @wozz. The client/connection classes are mimicked from their Python counterparts.

image

image

image

### DIFF
--- a/dashboard/.gitignore
+++ b/dashboard/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+dist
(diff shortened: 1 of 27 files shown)


marcelveldt (approved):

marcelveldt:
I'm going to merge this now, so we can prepare the build and inclusion of the UI on the webserver. 
We extend later in follow-up PR's

PR #548, opened 2024-02-14
Fix ruff low hanging fruit

marcelveldt:
Fix several lint issue sand optimizations based.
Mostly import order and new-style type notations.

### DIFF
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -4,23 +4,20 @@ from __future__ import annotations
 
 import asyncio
 import logging
-from types import TracebackType
-from typing import TYPE_CHECKING, Any, Callable, Dict, Final, Optional, cast
 import uuid
+from typing import TYPE_CHECKING, Any, Final, cast
 
-from aiohttp import ClientSession
 from chip.clusters import Objects as Clusters
 from chip.clusters.Types import NullValue
 
 from matter_server.common.errors import ERROR_MAP, NodeNotExists
-
-from ..common.helpers.util import (
+from matter_server.common.helpers.util import (
     convert_ip_address,
     convert_mac_address,
     dataclass_from_dict,
     dataclass_to_dict,
 )
-from ..common.models import (
+from matter_server.common.models import (
     APICommand,
     CommandMessage,
     CommissionableNodeData,
@@ -37,6 +34,7 @@ from ..common.models import (
     ServerInfoMessage,
     SuccessResultMessage,
 )
+
 from .connection import MatterClientConnection
 from .exceptions import ConnectionClosed, InvalidServerVersion, InvalidState
 from .models.node import (
@@ -48,6 +46,10 @@ from .models.node import (
 )
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+    from types import TracebackType
+
+    from aiohttp import ClientSession
     from chip.clusters.Objects import ClusterCommand
 
 SUB_WILDCARD: Final = "*"
@@ -58,12 +60,12 @@ SUB_WILDCARD: Final = "*"
 class MatterClient:
     """Manage a Matter server over WebSockets."""
 
-    def __init__(self, ws_server_url: str, aiohttp_session: ClientSession):
+    def __init__(self, ws_server_url: str, aiohttp_session: ClientSession) -> None:
         """Initialize the Client class."""
         self.connection = MatterClientConnection(ws_server_url, aiohttp_session)
         self.logger = logging.getLogger(__package__)
-        self._nodes: Dict[int, MatterNode] = {}
-        self._result_futures: Dict[str, asyncio.Future] = {}
+        self._nodes: dict[int, MatterNode] = {}
+        self._result_futures: dict[str, asyncio.Future] = {}
         self._subscribers: dict[str, list[Callable[[EventType, Any], None]]] = {}
         self._stop_called: bool = False
         self._loop: asyncio.AbstractEventLoop | None = None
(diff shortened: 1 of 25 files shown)


matter_server/client/client.py
```diff
@@ -48,6 +46,10 @@
 )
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
```
MartinHjelmare, line 49: I don't like this rule, that names that are only used for type checking should be imported under `if TYPE_CHECKING` flag. I think that's just annoying. I don't think there's big performance problem.

https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/

matter_server/client/client.py
```diff
@@ -125,7 +121,8 @@ def get_node(self, node_id: int) -> MatterNode:
         """Return Matter node by id or None if no node exists by that id."""
         if node := self._nodes.get(node_id):
             return node
-        raise NodeNotExists(f"Node {node_id} does not exist or is not yet interviewed")
+        msg = f"Node {node_id} does not exist or is not yet interviewed"
+        raise NodeNotExists(msg)
```
MartinHjelmare, line 125: I don't understand what makes it less readable in the traceback:
https://docs.astral.sh/ruff/rules/f-string-in-exception/

Not sure about this one.

marcelveldt:
This is not working as the linter config differs

matter_server/client/client.py
```diff
@@ -4,23 +4,20 @@
 
 import asyncio
 import logging
-from types import TracebackType
-from typing import TYPE_CHECKING, Any, Callable, Dict, Final, Optional, cast
 import uuid
+from typing import TYPE_CHECKING, Any, Final, cast
```
MartinHjelmare, line 8: I prefer having the imports sorted :abcd: regardless if they are imported with `import ...` or with `from ... import`. Can we adjust that rule? We have this sorted like that in core.

matter_server/server/client_handler.py
```diff
@@ -229,7 +231,7 @@ def _send_message(self, message: MessageType) -> None:
         try:
             self._to_write.put_nowait(_message)
         except asyncio.QueueFull:
-            self._logger.error(
+            self._logger.exception(
```
MartinHjelmare, line 234: Which rule is this?

scripts/beautify_diagnostics.py
```diff
@@ -2,33 +2,29 @@
...
         sys.exit(1)
 
     with open(sys.argv[1]) as f:
         data = json.load(f)
 
-    if "node" in data["data"]:
-        nodes = [data["data"]["node"]]
-    else:
-        nodes = data["data"]["server"]["nodes"]
+    nodes = [data["data"]["node"]] if "node" in data["data"] else data["data"]["server"]["nodes"]
```
MartinHjelmare, line 19: This creates a line that is longer than 88 characters. I don't like that.

scripts/generate_devices.py
```diff
@@ -83,21 +84,16 @@ def __hash__(self) -> int:
         name = device["typeName"]
 
         if not name.startswith("Matter "):
-            print("Unexpected: device doesn't start with Matter. Skipping")
```
MartinHjelmare, line 86: We don't need to remove prints from scripts if we think the prints are useful.

matter_server/client/client.py
marcelveldt: yeah I was surprised by it as well - we can disable this one if you want

matter_server/client/client.py
marcelveldt: @frenck do you know how ?

scripts/beautify_diagnostics.py
marcelveldt: I have set the minimal line length to 100 as 88 is really short

scripts/generate_devices.py
marcelveldt: yeah scripts/examples may have prints

matter_server/server/client_handler.py
marcelveldt: I don't know - its very hard to track

MartinHjelmare: I think it's this one:
https://docs.astral.sh/ruff/rules/error-instead-of-exception/

That's very subjective I'd say. Not sure that we should follow that one all the time.

matter_server/client/client.py
marcelveldt: We can disable it if we want ?

scripts/beautify_diagnostics.py
MartinHjelmare: I'd rather stick with 88 characters per line. That's what we have for all projects now. It also fits nicely on mobile in landscape.

matter_server/client/client.py
marcelveldt: @frenck @MartinHjelmare do we want this rule to be disabled by default ?

matter_server/client/client.py
marcelveldt: @frenck @MartinHjelmare do we want this rule to be disabled by default ?

matter_server/server/client_handler.py
marcelveldt: I've disabled this rule in the config

matter_server/client/client.py
MartinHjelmare: Personally I don't think this rule improves things.

marcelveldt: Me too, I'm in favor as well to just disable it

matter_server/client/client.py
marcelveldt: disabled in #567

PR #557, opened 2024-02-15
Run rollup as well before start development

agners:


agners:
Not sure if technically `rollup -c` is required too, since the later one watches the files and does the rollup stuff? It seems #554 is good enough to fix the issue on my side.

PR #566, opened 2024-02-20
Fix Nullable values in write attributes procedure

BDeliers:
Writing attributes values with JSON data is allowed in Matter, for instance with the ACL cluster we can send this payload:
<details>
<summary>Expand</summary>

```
[
    {
        "fabricIndex":0,
        "privilege":5,
        "authMode":2,
        "subjects":[
            112233
        ],
        "targets":null
    },
    {
        "fabricIndex":0,
        "privilege":1,
        "authMode":2,
        "subjects":[
            4444,
            5555,
            6666
        ],
        "targets":null
    },
    {
        "fabricIndex":0,
        "privilege":3,
        "authMode":3,
        "subjects":[
            123,
            456
        ],
        "targets":[
            {
                "cluster":6,
                "endpoint":null,
                "deviceType":null
            },
            {
                "cluster":null,
                "endpoint":1,
                "deviceType":null
            },
            {
                "cluster":8,
                "endpoint":2,
                "deviceType":null
            }
        ]
    }
]
```

</details>

When Python parses this JSON, the `null` value is then converted to a Python `None`. This `None` is not allowed by the Matter SDK, at the opposite of the `Nullable` type which is defined in the SDK to handle this kind of situation.

This pull-request is there to fix this issue: the Python `None` is converted to a Matter `Null` value of type `Nullable`.

### DIFF
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -3,17 +3,16 @@
 from __future__ import annotations
 
 import base64
-import binascii
-import logging
-import platform
-import socket
 from base64 import b64decode
+import binascii
 from dataclasses import MISSING, asdict, fields, is_dataclass
 from datetime import datetime
 from enum import Enum
 from functools import cache
-from importlib.metadata import PackageNotFoundError
-from importlib.metadata import version as pkg_version
+from importlib.metadata import PackageNotFoundError, version as pkg_version
+import logging
+import platform
+import socket
 from types import NoneType, UnionType
 from typing import (
     TYPE_CHECKING,
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -3,16 +3,17 @@
 from __future__ import annotations
 
 import base64
-from base64 import b64decode
 import binascii
+import logging
+import platform
+import socket
+from base64 import b64decode
 from dataclasses import MISSING, asdict, fields, is_dataclass
 from datetime import datetime
 from enum import Enum
 from functools import cache
-from importlib.metadata import PackageNotFoundError, version as pkg_version
-import logging
-import platform
-import socket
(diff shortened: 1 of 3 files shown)


marcelveldt:
First of all thanks for the PR!
I'm sorry but this is the wrong way to fix this - it needs to be addressed in our helper that reconstructs the Cluster object from the json in the api. 

It should be as easy as replacing the value assignment with our "parse_value" helper form utils.
That is already ready for the custom sdk types.

If you want I can make the adjustment and would be great if you can test if that gives the desired behavior

BDeliers:
Thanks for your feedback. If I understand properly what you said, it can be fixed by changing [Line 137 of matter_server/common/helpers/util.py](https://github.com/home-assistant-libs/python-matter-server/blob/main/matter_server/common/helpers/util.py#L137) from 
```
if value is None and value_type is Nullable:
        return None
```
to
```
if value is None and value_type is Nullable:
        return Null
```

?

marcelveldt:
Well, yes but we need some more. We will also need a boolean option to that util function if the custom types are allowed.
Because in this case you are going from [json --> dataclass convertor --> SDK] and then the custom SDK types are needed but when we do the same in e.g. Home Assistant or any other consumer of the lib we need to convert the values.

Want me to edit the PR to adjust the utils func so you can use it in the write command ?

BDeliers:
It seems that you know the project way better than I do.
Yes please, do the changes and I will then test with my use case !

marcelveldt:
I will do so monday. I'll let you know once its ready to test

marcelveldt:
I have modified your PR with the adjustments. Now all values that are sent to the SDK will be in the SDK's custom data types such as Nullable, uint and float32

marcelveldt:
I've tested both sending commands and write attribute and both still work.

agners (approved):

PR #567, opened 2024-02-21
A few small improvements to the linter config

marcelveldt:
Address a few small improvements to the linter config, as discussed in #548

- Disable Ruff TCH001 and TCH003 (requirement to move imports into type checking block)
- Disable ruff rule EM102 (disallow f strings in log)


MartinHjelmare (approved):
Thanks!

PR #568, opened 2024-02-21
Delay marking a node as unavailable

marcelveldt:
After our recent change with the proactive node detection using our own mdns discovery we also changed how/when we mark a node as unavailable/offline if its not responding. we lowered the treshold in favor of re-detecting it over mdns.

We however get multiple reports that people now see their nodes go unavailable a lot more often.
Now we can go argue that they should fix their networks/devices so they do not miss the subscription intervals but real world is that dropouts may happen and bad devices do exist ;-)

I have investigated other platforms like Apple Home and Google home and they're all a huge bit more forgiving than us, basically "hiding" the node unavailability for a while by debouncing it. This PR extends our tresholds as well to follow that pattern but it still logs these events so people are still aware that something is going on in their network (and hopefully fix it).

The outcome is that devices won't be marked unavailable in HA as fast as before, basically reproducing the behavior of other platforms.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -63,6 +63,8 @@ DATA_KEY_LAST_NODE_ID = "last_node_id"
 LOGGER = logging.getLogger(__name__)
 MAX_POLL_INTERVAL = 600
 MAX_COMMISSION_RETRIES = 3
+NODE_RESUBSCRIBE_ATTEMPTS_UNAVAILABLE = 3
+NODE_RESUBSCRIBE_TIMEOUT_OFFLINE = 30 * 60 * 1000
 
 MDNS_TYPE_OPERATIONAL_NODE = "_matter._tcp.local."
 MDNS_TYPE_COMMISSIONABLE_NODE = "_matterc._udp.local."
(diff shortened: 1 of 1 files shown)


marcelveldt:
I'm going to release a new beta based on this branch to allow some users to test it

matter_server/server/device_controller.py
```diff
@@ -976,14 +978,25 @@ def resubscription_attempted(
                 terminationError,
                 nextResubscribeIntervalMsec,
             )
+            resubscription_attempt = self._last_subscription_attempt[node_id] + 1
+            self._last_subscription_attempt[node_id] = resubscription_attempt
             # mark node as unavailable and signal consumers
             # we debounce it a bit so we only mark the node unavailable
```
MartinHjelmare, line 984:
```suggestion
            # Mark node as unavailable and signal consumers.
            # We debounce it a bit so we only mark the node unavailable
```

matter_server/server/device_controller.py
```diff
@@ -976,14 +978,25 @@ def resubscription_attempted(
...
-            # at the second resubscription attempt
-            if node.available and self._last_subscription_attempt[node_id] >= 1:
-                # NOTE: if the node is (re)discovered by mdns, that callback will
-                # take care of resubscribing to the node
+            # after some resubscription attempts and we shutdown the subscription
+            # if the resubscription interval exceeds 30 minutes (TTL of mdns)
+            # the node will be auto picked up by mdns if its alive again
```
MartinHjelmare, line 987:
```suggestion
            # after some resubscription attempts and we shutdown the subscription
            # if the resubscription interval exceeds 30 minutes (TTL of mdns).
            # The node will be auto picked up by mdns if it's alive again.
```

MartinHjelmare (approved):

PR #571, opened 2024-02-22
Skip push to main branch check in test workflow push

marcelveldt:
Our test workflow should ignore the pre-commit action for push to main if running from CI (but still run locally)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ name: Test
 
 on:
   push:
-    branches: [main]
+    branches: ["!main"]
   pull_request:
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ name: Test
 
 on:
   push:
-    branches: ["!main"]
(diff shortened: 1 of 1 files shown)


marcelveldt:
> I don't think we should do this. Tests should run on main branch too.

But then its duplicated ?
As it will run when you create the PR, run when you push to that PR's branch.

We never push to main directly (without a PR) because we have protected the branch as well as the pre-commit that prevents it.

marcelveldt:
I want to prevent this from happening:
https://github.com/home-assistant-libs/python-matter-server/actions/runs/7995064218/job/21834533096

MartinHjelmare:
This should work:

```
SKIP=no-commit-to-branch pre-commit run --all-files
```

marcelveldt:
> This should work:
> 
> ```
> SKIP=no-commit-to-branch pre-commit run --all-files
> ```

ah, and put that in an env variable ?

marcelveldt:
https://github.com/pre-commit/pre-commit/issues/1049

MartinHjelmare (approved):
Great!

PR #574, opened 2024-02-22
Extend sleep in release action

marcelveldt:
Turns out 60 seconds was still too short, extended it to 120 seconds.
Let's see next time if it succeeds.


marcelveldt:
build of the last beta went fine with the 120 seconds delay

MartinHjelmare (approved):

PR #585, opened 2024-02-26
Fix ping node and retrieval of IP addresses

marcelveldt:
Now that we have zeroconf in our server we can do a better job to get the known ip addresses.
This change makes sure that we fetch the (scoped) ip addresses by querying mdns and return cached info if the node is offline.

Also updated the node diagnostics in the client to use this new method of fetching the ip's if the server schema version matches, will fallback to the old method otherwise.

schema bumped to version 8 due to the new command but no breaking changes.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -693,7 +693,6 @@ class MatterDeviceController:
         async def _do_ping(ip_address: str) -> None:
             """Ping IP and add to result."""
             timeout = 10 if battery_powered else 2
-            clean_ip = ip_address.split("%")[0]
             if "%" in ip_address:
                 # ip address contains an interface index
                 clean_ip, interface_idx = ip_address.split("%", 1)
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -737,13 +737,7 @@ class MatterDeviceController:
         node_logger = LOGGER.getChild(f"[node {node_id}]")
         # query mdns for all IP's
         # ensure both fabric id and node id have 16 characters (prefix with zero's)
-        fabric_id_part = self.fabric_id_hex.upper()
-        while len(fabric_id_part) < 16:
-            fabric_id_part = "0" + fabric_id_part
-        node_id_part = f"{node_id:0x}"
-        while len(node_id_part) < 16:
-            node_id_part = "0" + node_id_part
-        mdns_name = f"{fabric_id_part}-{node_id_part}.{MDNS_TYPE_OPERATIONAL_NODE}"
+        mdns_name = f"{self.compressed_fabric_id:0{16}X}-{node_id:0{16}X}.{MDNS_TYPE_OPERATIONAL_NODE}"
         info = AsyncServiceInfo(MDNS_TYPE_OPERATIONAL_NODE, mdns_name)
         if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


matter_server/server/device_controller.py
```diff
@@ -696,41 +693,70 @@ async def ping_node(self, node_id: int) -> NodePingResult:
...
+        node_logger = LOGGER.getChild(f"[node {node_id}]")
+        # query mdns for all IP's
+        # ensure both fabric id and node id have 16 characters (prefix with zero's)
+        fabric_id_part = self.fabric_id_hex.upper()
+        while len(fabric_id_part) < 16:
+            fabric_id_part = "0" + fabric_id_part
+        node_id_part = f"{node_id:0x}"
+        while len(node_id_part) < 16:
+            node_id_part = "0" + node_id_part
+        mdns_name = f"{fabric_id_part}-{node_id_part}.{MDNS_TYPE_OPERATIONAL_NODE}"
```
agners, line 746: This can be simplified to:

```suggestion
        mdns_name = f"{self.compressed_fabric_id:0{16}X}-{node_id:0{16}X}.{MDNS_TYPE_OPERATIONAL_NODE}"
```

matter_server/server/device_controller.py
```diff
@@ -696,41 +693,70 @@ async def ping_node(self, node_id: int) -> NodePingResult:
...
-            )
-            # ignore invalid/non-operational interfaces
-            if not network_interface.isOperational:
-                continue
-            if network_interface.type in (
-                Clusters.GeneralDiagnostics.Enums.InterfaceTypeEnum.kUnspecified,
-                Clusters.GeneralDiagnostics.Enums.InterfaceTypeEnum.kUnknownEnumValue,
-            ):
-                continue
+            clean_ip = ip_address.split("%")[0]
```
agners, line 696: The scope ID (followed by the % sign) is important for link-local `fe80` addresses. Depending on the routing table, pinging a fe80 address might work or fail without scope ID.

Since scope ID remains valid on server side, and we get node addresses server side from a locally run mDNS, we should not remove it here but pass it to the ping command.

agners (changes requested):
Very nice update! I think checking the IP addresses we learn server side via mDNS is much more useful in practice :+1:

matter_server/server/device_controller.py
marcelveldt: I'll adjust it but I had some trouble with pytest choking on this syntax a while ago so I did it the manual way. I''' double check how it behaves now.

matter_server/server/device_controller.py
marcelveldt: that is exactly what we do. this clean ip is just there for logging

agners: Right, sorry I misread that :see_no_evil: 

Hm, but maybe we should also show the scope ID to the user? :thinking: 

We could also go for a compromise and log with scope id, but strip the scope ID from the result. What do you think?

marcelveldt: I think showing the scope to the user would make them think something is wrong. Let's go with logging it.

marcelveldt: adjusted the logging to include the interface (in a human readable way)

agners (approved):

Issue #558, opened 2024-02-16
AttributeError: 'NoneType' object has no attribute 'raise_on_error'


olavt:
I think there's a bug somewhere here:

2024-02-16 12:49:27 hemsedal-controller chip.DMG[1] ERROR Subscription Liveness timeout with SubscriptionID = 0x1f3b87dd, Peer = 01:0000000000000001
2024-02-16 12:49:27 hemsedal-controller matter_server.server.device_controller.[node 1][1] INFO Previous subscription failed with Error: 50, re-subscribing in 0 ms...
2024-02-16 12:49:37 hemsedal-controller chip.EM[1] ERROR Failed to Send CHIP MessageCounter:49887640 on exchange 48478i sendCount: 4 max retries: 4
2024-02-16 12:49:41 hemsedal-controller chip.SC[1] ERROR CASESession timed out while waiting for a response from the peer. Current state was 4
2024-02-16 12:49:41 hemsedal-controller chip.DMG[1] ERROR Failed to establish CASE for re-subscription with error 'src/protocols/secure_channel/CASESession.cpp:553: CHIP Error 0x00000032: Timeout'
2024-02-16 12:49:41 hemsedal-controller matter_server.server.device_controller.[node 1][1] INFO Previous subscription failed with Error: 50, re-subscribing in 7642 ms...
2024-02-16 12:49:41 hemsedal-controller matter_server.server.device_controller[1] INFO Marked node 1 as offline
2024-02-16 14:24:09 hemsedal-controller matter_server.server.device_controller[1] INFO Node 1 discovered on MDNS
2024-02-16 14:24:09 hemsedal-controller matter_server.server.device_controller.[node 1][1] INFO Setting up attributes and events subscription.
2024-02-16 14:24:09 hemsedal-controller matter_server.server.client_handler[1] ERROR [547320470672] Error handling message: CommandMessage(message_id='10', command='open_commissioning_window', args={'node_id': 1})
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/matter_server/server/client_handler.py", line 193, in _run_handler
    result = await result
             ^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/matter_server/server/device_controller.py", line 434, in open_commissioning_window
    sdk_result = await self._call_sdk(
                 ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #589, opened 2024-02-27
A few small fixes to the share (open commissioning window) feature

marcelveldt:
If a device has been put into commissioning mode and you issue the command again (e.g. by pressing the button in the HA interface) the device will throw a (rather cryptic) message because its already in commissioning mode.
Fix this by returning the previous parameters (and clear them once the timeout expires)

Closes https://github.com/home-assistant/core/issues/109936

Another improvement is creating a random discriminator instead of a hardcoded one.

Also, add a check if the node is actually available (so a case session is active) before attempting to open the commissioning window, which should prevent #558

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -414,7 +414,7 @@ class MatterDeviceController:
             return self._known_commissioning_params[node_id]
 
         if discriminator is None:
-            discriminator = randint(1000, 4096)  # noqa: S311
+            discriminator = randint(0, 4095)  # noqa: S311
 
         sdk_result = await self._call_sdk(
             self.chip_controller.OpenCommissioningWindow,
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -403,8 +405,16 @@ async def open_commissioning_window(
...
+            raise NodeNotReady(f"Node {node_id} is not (yet) available.")
+
+        if node_id in self._known_commissioning_params:
+            # node has already been put into commissioning mode,
+            # return previous parameters
+            return self._known_commissioning_params[node_id]
+
         if discriminator is None:
-            discriminator = 3840  # TODO generate random one
+            discriminator = randint(1000, 4096)  # noqa: S311
```
agners, line 417: Any reason to start from 1000? Also the constraint according to the OpenCommissioningWindow is to be between 0 and 4095, so:

```suggestion
            discriminator = randint(0, 4095)  # noqa: S311
```

agners (changes requested):
Hm, I see opening again doesn't extend the window. Storing the current credentials and return them is a nice service to the client here :+1:

marcelveldt:
> Hm, I see opening again doesn't extend the window. Storing the current credentials and return them is a nice service to the client here 👍

No, unfortunately not. You just have to wait until the current "sharing opening" expired before starting a new one.

agners (approved):

PR #593, opened 2024-02-27
Include (base) dashboard in the builtin-webserver

marcelveldt:
- Build the dashboard in the release script
- Host the dashboard on the builtin webserver
- Detect the production server based on our default port

This is just the base dashboard as-is with no added functionality, that is to be extended in follow-up PR's.

### DIFF
--- a/dashboard/src/entrypoint/main.ts
+++ b/dashboard/src/entrypoint/main.ts
@@ -10,7 +10,7 @@ async function main() {
     // Turn httpX url into wsX url and append "/ws"
     url = "ws" + new URL("./ws", location.href).toString().substring(4);
   } else {
-    // dvelopment server, ask for url to matter server
+    // development server, ask for url to matter server
     let storageUrl = localStorage.getItem("matterURL");
     if (!storageUrl) {
       storageUrl = prompt(
--- a/dashboard/src/entrypoint/main.ts
+++ b/dashboard/src/entrypoint/main.ts
@@ -3,16 +3,19 @@ import { MatterClient } from "../client/client";
 async function main() {
   import("../pages/matter-dashboard-app");
 
-  // Turn httpX url into wsX url and append "/ws"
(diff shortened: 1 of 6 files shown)


dashboard/src/entrypoint/main.ts
```diff
@@ -3,16 +3,19 @@ import { MatterClient } from "../client/client";
...
-  // Inside Home Assistant ingress, we will not prompt for the URL
-  if (!location.pathname.endsWith("/ingress")) {
+  let url = "";
+  // Detect if we're running in the (production) webserver included in the matter server or not.
+  if (location.href.includes(":5580")) {
+    // production server running inside the matter server
+    // Turn httpX url into wsX url and append "/ws"
+    url = "ws" + new URL("./ws", location.href).toString().substring(4);
+  } else {
+    // dvelopment server, ask for url to matter server
```
MartinHjelmare, line 13:
```suggestion
    // development server, ask for url to matter server
```

matter_server/server/server.py
```diff
@@ -106,7 +112,22 @@ async def start(self) -> None:
         await self.device_controller.start()
         await self.vendor_info.start()
         mount_websocket(self, "/ws")
-        self.app.router.add_route("GET", "/", self._handle_info)
+        self.app.router.add_route("GET", "/info", self._handle_info)
+
+        # Host dashboard if the prebuilt files are detected
+        if DASHBOARD_DIR_EXISTS:
+            self.logger.debug("Detected dashboard files on %s", str(DASHBOARD_DIR))
```
MartinHjelmare, line 119: We can store the string of the dashboard directory in a local variable since we need it multiple times.

MartinHjelmare (approved):

matter_server/server/server.py
marcelveldt: done

PR #594, opened 2024-02-27
Catch TimeoutError while retrieving certificates

marcelveldt:
Some users had reported that their Matter server didn't start and [while inspecting their logs](https://github.com/home-assistant/core/issues/111251#issuecomment-1966840622) I noticed the fetching of certificates from Github was raising a TimeoutError we weren't catching.

Still, not retrieving the certificates can cause issues when trying to commission new nodes but a not starting Matter server isn't very nice either.

NOTE: It's not possible to load these certificates while running. They must be fetched before starting the devicecontroller.

### DIFF
--- a/matter_server/server/helpers/paa_certificates.py
+++ b/matter_server/server/helpers/paa_certificates.py
@@ -106,7 +106,7 @@ async def fetch_dcl_certificates(
                     )
                     LAST_CERT_IDS.add(paa["subjectKeyId"])
                     fetch_count += 1
-    except ClientError as err:
+    except (ClientError, TimeoutError) as err:
         LOGGER.warning(
             "Fetching latest certificates failed: error %s", err, exc_info=err
         )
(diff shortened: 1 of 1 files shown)


MartinHjelmare (approved):

jph290:
> Some users had reported that their Matter server didn't start and [while inspecting their logs](https://github.com/home-assistant/core/issues/111251#issuecomment-1966840622) I noticed the fetching of certificates from Github was raising a TimeoutError we weren't catching.
> 
> Still, not retrieving the certificates can cause issues when trying to commission new nodes but a not starting Matter server isn't very nice either.
> 
> NOTE: It's not possible to load these certificates while running. They must be fetched before starting the devicecontroller.

The issue seemed to resolve itself luckily, but glad that the error will be handled more gracefully in the future.  Must have been a routing issue with my ISP.  Thanks for the help!

PR #595, opened 2024-02-28
Extend sleep in release action

marcelveldt:
The release action just failed again with a 120 seconds sleep (due to the released pypi package couldn't be found).
Extended it to 5 minutes now, let's see how that works out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -70,7 +70,7 @@ jobs:
           user: __token__
           password: ${{ secrets.PYPI_TOKEN }}
       - name: Wait for PyPI
-        run: sleep 120
+        run: sleep 300
   build-and-push-container-image:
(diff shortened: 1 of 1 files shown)


agners:
:see_no_evil:  we can't be the only ones having this no? I really wonder if there isn't a better way of doing this. :thinking:

agners (approved):

marcelveldt:
> 🙈 we can't be the only ones having this no? I really wonder if there isn't a better way of doing this. 🤔

I see this with all my projects that publish a package to pypi first first and then use that same package in a container build.
I think the only way to avoid is to build a wheel first and use that wheel in the container instead of the pypi package or do what we do now, wait for the pypi package to show up. Maybe we can do it in a loop or whatever or clear some cache first. ?

PR #597, opened 2024-02-28
A collection of small (stability) fixes and tweaks

marcelveldt:
After testing the 5.7.0 beta we collected a few tweaks to stabilize and enhance the server, especially for larger devices.
One important fix was to limit access to the SDK to a single thread with a single thread executorpool (already merged in separate PR). We also discovered that we were not catching unhandled exceptions.

This branch had a number of iterations while testing on users but in the end this is what's left;

- Enhance the logging config with some common tweaks such as showing the thread.
- Catch all unhandled exceptions in threads and asyncio tasks.
- Only print full stacktraces for exceptions if debug logging is enabled, otherwise just log a pretty-fied version.
- A few small tweaks to the mdns discovery and subscription ceilings

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -105,7 +105,6 @@ class MatterDeviceController:
         self.thread_credentials_set: bool = False
         self.compressed_fabric_id: int | None = None
         self._node_lock: dict[int, asyncio.Lock] = {}
-        self._node_setup_lock: asyncio.Semaphore = asyncio.Semaphore(5)
         self._aiobrowser: AsyncServiceBrowser | None = None
         self._aiozc: AsyncZeroconf | None = None
         self._sdk_executor = ThreadPoolExecutor(max_workers=1)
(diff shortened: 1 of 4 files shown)


agners (approved):

matter_server/server/device_controller.py
```diff
@@ -718,7 +724,7 @@ async def ping_node(self, node_id: int) -> NodePingResult:
 
         async def _do_ping(ip_address: str) -> None:
             """Ping IP and add to result."""
-            timeout = 10 if battery_powered else 2
+            timeout = 60 if battery_powered else 10
```
MartinHjelmare, line 727: Maybe create constants for these numbers to keep all time periods in one place?

matter_server/server/device_controller.py
```diff
@@ -1174,18 +1187,20 @@ def _on_mdns_operational_node_state(
         # mdns events for matter devices arrive in bursts of (duplicate) messages
         # so we debounce this as we only use the mdns messages for operational node discovery
         # and we have other logic in place to determine node aliveness
-
         now = time.time()
-        last_seen = self._mdns_last_seen.get(node_id, 0)
-        self._mdns_last_seen[node_id] = now
-        if now - last_seen < NODE_SUBSCRIPTION_CEILING:
+        last_seen = self._node_last_seen.get(node_id, 0)
+        if node.available and now - last_seen < 60:
```
MartinHjelmare, line 1192: And this number too.

MartinHjelmare (approved):

PR #599, opened 2024-02-28
Use single SDK worker Thread

agners:
Using multiple worker Thread when calling the SDK doesn't really add any benefit: The Python wrapper posts all SDK calls onto a single Thread anyways (see ChipStack.py:Call()). Instead, use a single Thread to call the Python wrapper.

This potentially avoids race condition in the Python wrapper.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import asyncio
 from collections import deque
+from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
 from functools import partial
 import logging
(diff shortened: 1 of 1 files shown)


marcelveldt:
I think the better way to do this is specify it in the __main__.py where we instantiate the loop and attach the executorpool.

agners:
I think ideally we should really call the SDK from a single Thread only. We could just use the default Thread executor pool, and set that one to 1. But that means that in case we want other operation (non-SDK related ones) use that single Thread too.

In the end, either way works for me. Ultimately it would be better if we have the whole Python wrapper to be asyncio native :smile:

marcelveldt:
Let's go with this for now but in the long run we should have a single SDK specific executor specified in the stack and then use that executor from all places that communicate with the SDK, including the device controller.

PR #604, opened 2024-02-29
Add timeouts to sdk calls and subscription set-up

marcelveldt:
While skimming through user logs I noticed that sometimes a call to the sdk (especially the read attribute one) seems to be hanging forever. I think its good for detection and deadlock-preventing to add timeouts to the sdk calls.

I've set the default to 5 minutes per call. Now with a single executor thread handling all calls to the SDK that seems reasonable. If calls take (even) longer than 5 minutes to complete, something is seriously wrong.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1090,18 +1090,27 @@ class MatterDeviceController:
             ):
                 try:
                     await self.interview_node(node_id)
-                except (NodeNotResolving, NodeInterviewFailed, TimeoutError) as err:
-                    LOGGER.warning("Unable to interview Node %s", exc_info=err)
+                except (NodeNotResolving, NodeInterviewFailed) as err:
+                    LOGGER.warning(
+                        "Unable to interview Node %s: %s",
+                        node_id,
+                        str(err),
+                        # log full stack trace if debug logging is enabled
+                        exc_info=err if LOGGER.isEnabledFor(logging.DEBUG) else None,
+                    )
                     # NOTE: the node will be picked up by mdns discovery automatically
                     # when it comes available again.
                     return
             # setup subscriptions for the node
             try:
                 await self._subscribe_node(node_id)
-            except (NodeNotResolving, TimeoutError):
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -1076,15 +1090,15 @@ async def _setup_node(self, node_id: int) -> None:
             ):
                 try:
                     await self.interview_node(node_id)
-                except (NodeNotResolving, NodeInterviewFailed) as err:
+                except (NodeNotResolving, NodeInterviewFailed, TimeoutError) as err:
```
agners, line 1093: Can you log Timeout errors separately here..

matter_server/server/device_controller.py
```diff
@@ -1076,15 +1090,15 @@ async def _setup_node(self, node_id: int) -> None:
...
                     LOGGER.warning("Unable to interview Node %s", exc_info=err)
                     # NOTE: the node will be picked up by mdns discovery automatically
                     # when it comes available again.
                     return
             # setup subscriptions for the node
             try:
                 await self._subscribe_node(node_id)
-            except NodeNotResolving:
+            except (NodeNotResolving, TimeoutError):
```
agners, line 1101: ...and here?

This way we know if this was aborted due to timeout or some other problem,

agners (commented):
In general, the SDK should have it's own timeout. When we abort here on Matter Server side, the call will continue anyways... That said, I am not against adding this. It might help to find and analyze such long running calls. We just need to make sure that we also can identify those as this.

matter_server/server/device_controller.py
marcelveldt: it is being logged as the exception is sent to the logger

agners: Right :see_no_evil: . So this case is fine! :+1:

matter_server/server/device_controller.py
marcelveldt: added additional logging info

agners (approved):

PR #609, opened 2024-02-29
Add option to skip downloading PAA root certificates

agners:
This is especially useful during development.

### DIFF
--- a/matter_server/server/__main__.py
+++ b/matter_server/server/__main__.py
@@ -91,6 +91,12 @@ parser.add_argument(
     default=None,
     help="Primary network interface for link-local addresses (optional).",
 )
+parser.add_argument(
+    "--skip-paa-certificate-fetch",
+    dest="fetch_paa",
+    action="store_false",
+    help="Do not fetch PAA root certificates on startup.",
+)
 
 args = parser.parse_args()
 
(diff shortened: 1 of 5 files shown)


agners:
I restart the server often in my dev environment, so I'd like to have the option to skip (as I know they are recent enough).

marcelveldt:
> I restart the server often in my dev environment, so I'd like to have the option to skip (as I know they are recent enough, and it slows down startup time of the Matter Server, which is annoying during testing). This allows me to just add `--skip-paa-certificate-fetch` to the VS Code task run configuration.

Yeah, I get the point, but maybe we can enhance this even further by storing the last fetch timestamp in our persistent storage and if its fresher than, let's say 24 hours, we skip downloading the certs. That way its even more convenient.

agners:
> Yeah, I get the point, but maybe we can enhance this even further by storing the last fetch timestamp in our persistent storage and if its fresher than, let's say 24 hours, we skip downloading the certs. That way its even more convenient.

Good idea, also for production!

But we could still have this flag, IMHO :sweat_smile:

marcelveldt:
> But we could still have this flag, IMHO 😅

Let's not do that. It will end up by default in our VS Code configs and then we get issues one day when we try to commission a brand new device and then realize only after 3 hours we have this flag set haha. I'd rather quickly implement that freshness check. Then its solved for both usecases.

agners:
Yeah I guess this is philosophical. I kinda like to have arguments and to be under full control over what my server is doing. :man_shrugging: 

But I can live with it either way.

PR #611, opened 2024-03-01
Fetch certificates only once a day

agners:
Skip fetching the certificates if they have been fetched within the last 24h already.


marcelveldt:
haha you went creative, I was thinking to just store a timestamp in the persistent storage but this is also good.

marcelveldt:
Note that we are still compatible with python3.10 and that is probably the reason that the UTC lookup fails

agners:
> haha you went creative, I was thinking to just store a timestamp in the persistent storage but this is also good.

I mean, this is storing the timestamp where it belongs: Close to the data we try to timestamp. And use of file system metadata :sweat_smile:

marcelveldt:
> > haha you went creative, I was thinking to just store a timestamp in the persistent storage but this is also good.
> 
> I mean, this is storing the timestamp where it belongs: Close to the data we try to timestamp. And use of file system metadata 😅

Yeah, this is even better than a simple timestamp in our storage file so well done.

marcelveldt (approved):

PR #618, opened 2024-03-05
Several improvements to resolving and (re)subscribing to nodes

marcelveldt:
- Remove Timeout (as it is no longer needed now we identified the source)
- Allow cached ip in background scanner
- Use SDK resolve in node setup instead of zeroconf/ping
- Allow running the loop in debug mode on dev setup
- fix interview_node
- handle race condition in ping
- better debouncing of mdns

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -501,7 +501,6 @@ class MatterDeviceController:
 
         try:
             if not (node := self._nodes.get(node_id)) or not node.available:
-                await self._resolve_node(node_id=node_id)
                 LOGGER.info("Interviewing node: %s", node_id)
                 read_response: Attribute.AsyncReadTransaction.ReadResponse = (
                     await self.chip_controller.Read(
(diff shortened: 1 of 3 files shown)


matter_server/server/server.py
```diff
@@ -133,6 +133,7 @@ async def start(self) -> None:
             )
         self.loop = asyncio.get_running_loop()
         self.loop.set_exception_handler(_global_loop_exception_handler)
+        self.loop.set_debug(os.environ.get("PYTHONDEBUG") == "1")
```
agners, line 136: Nit: Definition of this environment variable seems to be debugging is enabled if this string is "non-empty" (https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDEBUG), so 

```suggestion
        self.loop.set_debug(os.environ.get("PYTHONDEBUG") != "")
```

matter_server/server/server.py
```diff
@@ -219,11 +220,13 @@ def get_diagnostics(self) -> ServerDiagnostics:
...
     def signal_event(self, evt: EventType, data: Any = None) -> None:
         """Signal event to listeners."""
+        if TYPE_CHECKING:
+            assert self.loop
         for callback in self._subscribers:
             if asyncio.iscoroutinefunction(callback):
                 asyncio.create_task(callback(evt, data))
             else:
-                callback(evt, data)
+                self.loop.call_soon_threadsafe(callback, evt, data)
```
agners, line 229: Hm, I guess technically this is not needed in every `signal_event` call but maybe better safe than sorry :thinking:

matter_server/server/device_controller.py
```diff
@@ -682,14 +682,12 @@ async def remove_node(self, node_id: int) -> None:
...
         try:
-            result = await self._call_sdk(
-                self.chip_controller.SendCommand(
-                    nodeid=node_id,
-                    endpoint=0,
-                    payload=Clusters.OperationalCredentials.Commands.RemoveFabric(
-                        fabricIndex=fabric_index,
-                    ),
-                )
+            result = await self.chip_controller.SendCommand(
```
agners, line 685: So is SendCommand safe to call from a different Thread?

matter_server/server/server.py
marcelveldt: well then we also need to check for None

matter_server/server/device_controller.py
marcelveldt: Well, that is the big mystery for all these async calls. My current hunch is that it is as its called from the mainthread.

matter_server/server/server.py
marcelveldt: Corrected the code to assig a default of empty string

matter_server/server/device_controller.py
agners: Ah I see this is an async one. I mean we call it from (our) main thread. The SendCommand then uses `src/controller/python/chip/ChipStack.py:ChipStack Call`, it uses thread safe locks (the networkLock) and `PostTaskOnChipThread` to run the task on the chip Thread. So I think it is fairly safe to assume that we can run this from any Thread. :+1:

agners (approved):

PR #623, opened 2024-03-07
Add warning if node setup takes too long

marcelveldt:
Last couple of weeks we have been investigating some sdk lockup issues. In fact it's not a complete deadlock but a node setup (initial read subscription) that takes forever. In our observations the transmission is halfway aborted in the read request and then resumed in some sort of loop, so the Read request never finishes.

It's hard to reproduce the issue and if its limited to only Nanoleaf bulbs or also other devices and if the type of Border routers make a difference. 

This PR adds a big fat warning in the log if we detect this situation if the node setup takes longer than 15 minutes.
It will keep logging this message every 15 minutes until the user takes action. There are only 5 slots available for node setup so if all 5 are taken, the server will stall setting up any new nodes.

The issue should not be ignored by the user but action taken, either by resolving the connectivity issue(s) or by reporting this to us so we can keep track.

example of the warning in the log:

Scherm­afbeelding 2024-03-07 om 18 08 38

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1067,8 +1067,42 @@ class MatterDeviceController:
             return
         self._nodes_in_setup.add(node_id)
         node_logger = LOGGER.getChild(f"node_{node_id}")
-        try:
-            async with self._node_setup_throttle:
+        node_data = self._nodes[node_id]
+        log_timers: dict[int, asyncio.TimerHandle] = {}
+
+        def log_node_long_setup(time_start: float) -> None:
+            """Temporary measure to track a locked-up SDK issue in some (special) circumstances."""
+            time_mins = int((time.time() - time_start) / 60)
+            node_model = node_data.attributes.get(
+                "0/40/14", node_data.attributes.get("0/40/3", "")
+            )
+            node_name = f"Node {node_id} ({node_model})"
+
+            node_logger.error(
+                f"\n\nATTENTION: {node_name} did not complete setup in {time_mins} minutes.\n"  # noqa: G004
+                "This is an indication of a (connectivity) issue with this device. \n"
+                "Try powercycling this device and/or relocate it closer to a Border Router or \n"
+                "WiFi Access Point. If this issue persists, please create an issue report on \n"
+                "the Matter channel of the Home Assistant Discord server or on Github:\n"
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -1067,8 +1067,42 @@ async def _setup_node(self, node_id: int) -> None:
...
-        try:
-            async with self._node_setup_throttle:
+        node_data = self._nodes[node_id]
+        log_timers: dict[int, asyncio.TimerHandle] = {}
+
+        def log_node_long_setup(time_start: float) -> None:
+            """Temporary measure to track a locked-up SDK issue in some (special) circumstances."""
+            time_mins = int((time.time() - time_start) / 60)
+            node_model = node_data.attributes.get(
+                "0/40/14", node_data.attributes.get("0/40/3", "")
```
agners, line 1077: Can we add a comment what these numbers mean exactly?

matter_server/server/device_controller.py
```diff
@@ -1067,8 +1067,42 @@ async def _setup_node(self, node_id: int) -> None:
...
+            node_name = f"Node {node_id} ({node_model})"
+
+            node_logger.error(
+                f"\n\nATTENTION: {node_name} did not complete setup in {time_mins} minutes.\n"  # noqa: G004
+                "This is an indication of a (connectivity) issue with this device. \n"
+                "Try powercycling this device and/or relocate it closer to a Border Router or \n"
+                "WiFi Access Point. If this issue persists, please create an issue report on \n"
+                "the Matter channel of the Home Assistant Discord server or on Github:\n"
+                "https://github.com/home-assistant/core/issues/new?assignees=&labels="
+                "integration%3A%20matter&projects=&template=bug_report.yml\n",
```
agners, line 1088: Can we also print the (last known) IP here maybe? :thinking:  Might be helpful for debugging.

matter_server/server/device_controller.py
marcelveldt: Its the modelname from the raw attributes ;-)

matter_server/server/device_controller.py
marcelveldt: added

matter_server/server/device_controller.py
marcelveldt: added comment

agners (approved):

PR #648, opened 2024-03-27
Make subscription attribute update handler Thread safe

agners:
The attribute update callback of subscription is called from the Matter SDK Thread directly. Most handling needs to run in the event loop Thread. Do a few quick checks, and then call the rest of the handler in the event loop Thread.

This also gets rid of a couple of `loop.create_task()` calls from the Matter SDK thread. `loop.create_task()` is not Thread safe.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -847,7 +847,55 @@ class MatterDeviceController:
 
         loop = cast(asyncio.AbstractEventLoop, self.server.loop)
 
-        # set-up the actual subscription
+        def attribute_updated(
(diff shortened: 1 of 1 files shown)


agners:
Instead of calling a routine on the event loop thread, we could also call a co-routine directly using `asyncio.run_coroutine_threadsafe`. Then we *might* also be able to remove some of the `create_task` calls and just await those calls. But this could affect timing etc. In a first step, I'd like to just make this Thread safe while leaving things mostly as is.

balloob (approved):
Sweet

PR #651, opened 2024-03-28
Add support for importing (temporary) test nodes

marcelveldt:
This adds support to both the backend, client and the dashboard to import test nodes from a HA diagnostics dump.
This makes it easier to debug (device)specific issues and extending device type support.

- You can import a diagnostics file from the builtin dashboard
- It takes both a full (multi node) dump from the config entry and a single node dump
- the test nodes are also available in the client so you can debug it in HA
- actions performed on one of the test nodes are verbosely logged
- removing test nodes can be done by a regular remove command or restarting the server
- testnodes are not persisted on storage, they're memory only.

### DIFF
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -306,8 +306,8 @@ class MatterClient:
         mac_address = None
         attribute = Clusters.GeneralDiagnostics.Attributes.NetworkInterfaces
         network_interface: Clusters.GeneralDiagnostics.Structs.NetworkInterface
-        for network_interface in node.get_attribute_value(
-            0, cluster=None, attribute=attribute
+        for network_interface in (
+            node.get_attribute_value(0, cluster=None, attribute=attribute) or []
         ):
             # ignore invalid/non-operational interfaces
             if not network_interface.isOperational:
@@ -332,6 +332,11 @@ class MatterClient:
                 continue
             mac_address = convert_mac_address(network_interface.hardwareAddress)
             break
+        else:
+            self.logger.warning(
+                "Could not determine network_interface info for Node %s, "
+                "is it missing the GeneralDiagnostics/NetworkInterfaces Attribute?"
+            )
         # get thread/wifi specific info
         node_type = NodeType.UNKNOWN
         network_name = None
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -29,6 +29,7 @@ from matter_server.server.helpers.attributes import parse_attributes_from_read_r
 from matter_server.server.helpers.utils import ping_ip
 
 from ..common.errors import (
+    InvalidArguments,
     NodeCommissionFailed,
     NodeInterviewFailed,
     NodeNotExists,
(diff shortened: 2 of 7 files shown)


matter_server/server/device_controller.py
```diff
@@ -567,6 +578,18 @@ async def send_device_command(
         cluster_cls: Cluster = ALL_CLUSTERS[cluster_id]
         command_cls = getattr(cluster_cls.Commands, command_name)
         command = dataclass_from_dict(command_cls, payload, allow_sdk_types=True)
+        if node_id >= TEST_NODE_START:
+            LOGGER.info(
```
MartinHjelmare, line 582: Aren't these logging messages more appropriate at debug level?

matter_server/server/device_controller.py
```diff
@@ -819,6 +866,28 @@ async def get_node_ip_addresses(
...
+    @api_command(APICommand.IMPORT_TEST_NODE)
+    async def import_test_node(self, dump: str) -> None:
+        """Import test node(s) from a HA or Matter server diagnostics dump."""
+        try:
+            dump_data = cast(dict, json_loads(dump))
+        except JSON_DECODE_EXCEPTIONS as err:
+            raise InvalidArguments("Invalid json") from err
+        # dump can either be a single dump or a full dump with multiple nodes
+        dump_nodes: list[dict[str, Any]]
+        if "node" in dump_data["data"]:
```
MartinHjelmare, line 878: It would be good to document the dump schema that needs to be followed here.

matter_server/server/device_controller.py
marcelveldt: Well it depends on the usecase. In this case the logging is a replacement of the action and not logging the action, so I figured INFO more useful.

This just logs all actions performed on a test node instead of actually performing it.

Just some thoughts. I can change it to debug, no problem.

matter_server/server/device_controller.py
marcelveldt: yeah its HA diagnostics but I figured we might also want to create an export from the dashboard itself

marcelveldt: I have added a comment here about the format.
In a follow-up PR I'll add an export button in the dashboard as well to dump the diagnostics.

MartinHjelmare (approved):
Python looks ok. :+1:

matter_server/client/client.py
```diff
@@ -332,6 +332,11 @@ async def node_diagnostics(self, node_id: int) -> NodeDiagnostics:
                 continue
             mac_address = convert_mac_address(network_interface.hardwareAddress)
             break
+        else:
+            self.logger.warning(
+                "Could not determine network_interface info for Node %s, "
+                "is it missing the GeneralDiagnostics/NetworkInterfaces Attribute?"
+            )
```
MartinHjelmare, line 339: We're missing a log parameter for the node id.

MartinHjelmare (changes requested):
Noticed one issue.

matter_server/client/client.py
marcelveldt: fixed

matter_server/server/device_controller.py
marcelveldt: changed to debug (and opened PR to add verbose debug level)

MartinHjelmare (approved):

PR #653, opened 2024-03-28
Run write node state in executor

marcelveldt:
While on debugging spree I noticed asyncio debug mode warning me a couple of times that the "_write_node_state" callback was taking more than 100ms to complete. 

That is not a super big surprise due to all the parsing done to transform the dataclass into a dict in combination with a node with lots of attributes.  

Better be safe than sorry so wrapped it in an executor.

### DIFF
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import base64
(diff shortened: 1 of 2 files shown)


marcelveldt:
I just realized we dont have any custom types anymore in the storage file so we can directly dump to orjson

marcelveldt:
better solution in #658

PR #669, opened 2024-04-17
Expand dashboard to explore Node values

marcelveldt:
- Add script that generates human readable descriptions from all Matter device types, Clusters and Attributes
- Auto generate typescript models for the dashboard for the descriptions.
- Split up the dashboard components a bit into smaller chunks
- Show basic Node details
- Implement Node explorer  to browse endpoints, clusters and attributes including their values
- I did not yet  implement a way to change/write attributes (or commands) - its read only for now

- Exploring the node's clusters and attributes also works for virtual node, imported from a diagnostics dump which is super useful for debugging.

Scherm­afbeelding 2024-04-17 om 15 26 43

Scherm­afbeelding 2024-04-17 om 15 28 03

Scherm­afbeelding 2024-04-17 om 15 28 24


scripts/generate_descriptions.py
```diff
@@ -0,0 +1,116 @@
+"""Generate descriptions.json for the dashboard."""
+
+import pathlib  # noqa: I001
```
MartinHjelmare, line 3: Why is it complaining about unsorted imports and why can't we just sort it?

MartinHjelmare (commented):
Someone else should review the typescript code.

scripts/generate_descriptions.py
marcelveldt: fixed

dashboard/src/client/client.ts
```diff
@@ -188,7 +186,24 @@ export class MatterClient {
...
+
+    if (event.event === "node_updated") {
+      const node = new MatterNode(event.data);
+      this.nodes = { ...this.nodes, [node.node_id]: node };
+      this.fireEvent("nodes_changed");
+      return;
+    }
+
+    if (event.event === "attribute_updated") {
+      const nodeId = (event.data as Array<any>)[0] as number;
```
bramkragten, line 201: Why isn't `data` typed as array for the `ServerEventAttributeUpdated` event? It is now typed as an object?

marcelveldt: because it can contain anything, it is literally any

dashboard/src/client/client.ts
```diff
@@ -188,7 +186,24 @@ export class MatterClient {
...
+      const attributeKey = (event.data as Array<any>)[1] as string;
+      const attributeValue = (event.data as Array<any>)[2] as any;
+      const node = new MatterNode(this.nodes[nodeId]);
```
bramkragten, line 204: Isn't `this.nodes[nodeId]` already an instance of `MatterNode`? As you do save it back to `this.nodes` below?

dashboard/src/client/client.ts
marcelveldt: ah but wait... for the attribute_updated_event its actually always a tuple

dashboard/src/client/client.ts
marcelveldt: It was the only way I could get reactivity to respond but tell me if you have better ideas

dashboard/src/pages/components/header.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+@customElement("dashboard-header")
+export class DashboardHeader extends LitElement {
+
+  @property() public backButton?: string;
+  @property() public actions?: HeaderAction[];
+
+  protected render() {
+    const isProductionServer = location.href.includes(":5580") || location.href.includes("hassio_ingress");
```
bramkragten, line 25: This seems like a dirty solution? Also this should not be checked in every render, but only 1 time.

dashboard/src/pages/components/header.ts
```diff
@@ -0,0 +1,89 @@
...
+            <ha-svg-icon .path=${mdiArrowLeft}></ha-svg-icon>
+          </md-icon-button>
+        </a>` : ''}
+
+      <div>${this.title || ''}</div>
+      <div class="flex"></div>
+      <div class="actions">
+      ${this.actions?.map((action) => {
+      return html`
+          <md-icon-button @click=${action.action} title=${action.label}>
```
bramkragten, line 42:
```suggestion
          <md-icon-button @click=${action.action} .title=${action.label}>
```

dashboard/src/pages/components/header.ts
marcelveldt: yeah, I have no other way to check it for now. I'll make it an attribute of the client.

dashboard/src/pages/components/header.ts
```diff
@@ -0,0 +1,89 @@
...
+          </md-icon-button>
+          `}
+      </div>
+    </div>
+    `;
+  }
+
+  private _disconnect() {
+    localStorage.removeItem("matterURL");
+    location.reload();
```
bramkragten, line 62: Make this a helper function, on the same place the `connect` logic is.

dashboard/src/pages/components/node-details.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  protected render() {
+    if (!this.node) return html``;
+    return html`
+      <md-list>
+        <md-list-item>
+            <div slot="headline">
+                <b>Node ${this.node.node_id} ${this.node.nodeLabel}</b>
+                ${this.node.available
+        ? ""
```
bramkragten, line 49:
```suggestion
        ? nothing
```

dashboard/src/pages/components/server-details.ts
```diff
@@ -0,0 +1,123 @@
...
+@customElement("server-details")
+export class ServerDetails extends LitElement {
+
+  public client?: MatterClient;
+
+  protected render() {
+    if (!this.client) return html``;
+
+    const isProductionServer = location.href.includes(":5580") || location.href.includes("hassio_ingress");
+    const serverAddressBase = this.client.url.split("://")[1].split(":")[0];
```
bramkragten, line 28: These should probably not be in render

dashboard/src/pages/components/node-details.ts
```diff
@@ -0,0 +1,155 @@
...
+        title: "Remove",
+        text: "Are you sure you want to remove this node?",
+        confirmText: "Remove",
+      }))
+    ) {
+      return;
+    }
+    try {
+      await this.client.removeNode(this.node!.node_id);
+      location.replace("#");
```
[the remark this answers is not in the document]
marcelveldt: I think that is old code that can be removed as I fixed the reactivity issues.

dashboard/src/pages/components/server-details.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (
+      !(await showPromptDialog(this, {
+        title: "Add test node",
+        text: "Do you want to add a test node from a diagnostics dump ?",
+        confirmText: "Select file",
+      }))
+    ) {
+      return;
+    }
+    const fileElem = this.shadowRoot!.getElementById('fileElem') as HTMLInputElement;
```
bramkragten, line 82:
```suggestion
    const fileElem = this.renderRoot.getElementById('fileElem') as HTMLInputElement;
```

dashboard/src/pages/components/server-details.ts
```diff
@@ -0,0 +1,123 @@
...
+            showAlertDialog(this, {
+              title: "Failed to import test node",
+              text: err.message,
+            });
+          }
+        }
+      }
+      event.preventDefault();
+    }
+    fileElem!.addEventListener('change', handleInput);
```
bramkragten, line 103: We do we add and removed the event listener, and not just keep it on the element (and set it in the template)?

bramkragten, line 103: We do we add and removed the event listener, and not just keep it on the element (and set it in the template)?

dashboard/src/pages/matter-cluster-view.ts
```diff
@@ -0,0 +1,149 @@
...
+function clusterAttributes(attributes: { [key: string]: any }, endpoint: number, cluster: number) {
+  // extract unique clusters from the node attributes, as (sorted) array
+  const endpointClusters = Object.keys(attributes)
+    .filter(key => key.startsWith(`${endpoint}/${cluster}`))
+    .map((key) => {
+      const attributeKey = Number(key.split('/')[2]);
+      return { key: attributeKey, value: attributes[key] };
+    }, []);
+  return endpointClusters;
+}
```
bramkragten, line 29:
```suggestion
function clusterAttributes(attributes: { [key: string]: any }, endpoint: number, cluster: number) {
  // extract unique clusters from the node attributes, as (sorted) array
  return Object.keys(attributes)
    .filter(key => key.startsWith(`${endpoint}/${cluster}`))
    .map((key) => {
      const attributeKey = Number(key.split('/')[2]);
      return { key: attributeKey, value: attributes[key] };
    }, []);
}
```

dashboard/src/pages/matter-cluster-view.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  render() {
+
+    if (!this.node || this.endpoint == undefined || this.cluster == undefined) {
+      return html`
+        <p>Node, endpoint or cluster not found!</p>
+        <button
+          @click=${() => {
+          history.back();
+        }}
```
bramkragten, line 52: Dont define event handlers in your template, this means they will be recreated on every render

dashboard/src/pages/matter-endpoint-view.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  render() {
+
+    if (!this.node || this.endpoint == undefined) {
+      return html`
+        <p>Node or endpoint not found!</p>
+        <button
+          @click=${() => {
+          history.back();
+        }}
```
bramkragten, line 51: Same, dont inline

dashboard/src/pages/matter-server-view.ts
```diff
@@ -15,58 +23,93 @@ declare global {
...
+          ${nodes.map(([id, node]) => {
+      return html`
+              <md-list-item type="link" href=${`#node/${node.node_id}`}>
+                <div slot="headline">
+                Node ${node.node_id}
+                  ${node.available
+          ? ""
+          : html`<span class="status">OFFLINE</span>`}
+                </div>
+                <div slot="supporting-text">${node.nodeLabel ? `${node.nodeLabel} | ` : ''} ${node.vendorName} | ${node.productName}</div>
```
bramkragten, line 66:
```suggestion
                <div slot="supporting-text">${node.nodeLabel ? `${node.nodeLabel} | ` : nothing} ${node.vendorName} | ${node.productName}</div>
```

dashboard/src/client/client.ts
bramkragten: If it is can be anything your code is wrong, as you should check if it is an array first before assuming :-)

dashboard/src/client/client.ts
bramkragten: `nodes` is the reactive part right? You already make a copy of that

dashboard/src/pages/matter-endpoint-view.ts
marcelveldt: fixed

dashboard/src/pages/matter-cluster-view.ts
marcelveldt: fixed

dashboard/src/pages/components/node-details.ts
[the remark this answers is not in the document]
marcelveldt: fixed

dashboard/src/pages/components/server-details.ts
marcelveldt: fixed

dashboard/src/pages/components/header.ts
marcelveldt: done

dashboard/src/pages/components/node-details.ts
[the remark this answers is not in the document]
marcelveldt: Ah I remember it... if you remove a node from the node details page, to redirect you back to the root page

dashboard/src/pages/components/server-details.ts
marcelveldt: fixed

bramkragten (approved):

PR #676, opened 2024-04-24
Some small tweaks to the subscripions logic

marcelveldt:
While we're still hunting down the cause for the continuous "subscription failed with Error: 50, re-subscribing in 0 ms" errors where the subscription of a node suddenly gets lost with no reason, I found a few tweaks. Not really solutions to the issue but still good to apply.

- Increase subscription ceiling for WiFi devices to 60 seconds (same as mains powered thread devices) as we see the error mostly for WiFi devices.

- Remove keepSubscriptions on the subscription itself as its kind of redundant with the autoResubscribe and saves a bit of stress on the device itself (hopefully) - though this is still very unclear.

- Remove fabricFiltered=False (so reinstate it being set to True) on the subscription itself as the only reason we were doing that is to have an up to date list of fabrics at all times. Removing it in the sub saves a little bit of stress on the device.

- Refresh the node's fabrics in the diagnostics itself to ensure we have the latest info as an alternative for having the fabricFiltered=False in the sub.

### DIFF
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -18,6 +18,7 @@ from matter_server.common.errors import ERROR_MAP, NodeNotExists
 from ..common.helpers.util import (
     convert_ip_address,
     convert_mac_address,
+    create_attribute_path_from_attribute,
     dataclass_from_dict,
     dataclass_to_dict,
 )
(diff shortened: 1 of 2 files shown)


marcelveldt:
CI is failing but that is a global issue (signing issue on the ubuntu used for CI)

PR #684, opened 2024-05-01
Handle value specified as bytes received as integer

marcelveldt:
Somehow some devices send some values (such as the Noc) as integer instead of bytes so our parser chokes on that.
Not sure if this is device related or somehow the tlv parser doesnt handle it but its not super big deal.
Convert it so parsing the values doesnt crash in the client.

Fixes https://github.com/home-assistant/core/issues/116304
Fixes https://github.com/home-assistant/core/issues/113279

### DIFF
--- a/tests/common/test_parser.py
+++ b/tests/common/test_parser.py
@@ -107,6 +107,6 @@ def test_dataclass_from_dict():
     # test extra keys not silently ignored in strict mode
     with pytest.raises(KeyError):
         dataclass_from_dict(BasicModel, raw2, strict=True)
-    # test int gets converted to bytes
-    res = parse_value("int_value_that_should_be_bytes", 5, bytes)
-    assert res == b"\x00\x00\x00\x05"
+    # test NOCStruct.noc edge case
+    res = parse_value("NOCStruct.noc", 5, bytes)
+    assert res == b""
--- a/tests/common/test_parser.py
+++ b/tests/common/test_parser.py
@@ -77,7 +77,6 @@ def test_dataclass_from_dict():
     raw["b"] = 2
     res = dataclass_from_dict(BasicModel, raw)
     assert res.b == 2.0
-
     # test datetime string
     assert isinstance(res.f, datetime.datetime)
     assert res.f.month == 12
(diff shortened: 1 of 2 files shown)


tests/common/test_parser.py
```diff
@@ -107,3 +108,6 @@ def test_dataclass_from_dict():
     # test extra keys not silently ignored in strict mode
     with pytest.raises(KeyError):
         dataclass_from_dict(BasicModel, raw2, strict=True)
+    # test int gets converted to bytes
+    res = parse_value("int_value_that_should_be_bytes", 5, bytes)
+    assert isinstance(res, bytes)
```
MartinHjelmare, line 113: Maybe assert what the value is instead?

MartinHjelmare (approved):

tests/common/test_parser.py
marcelveldt: Yeah and I just realize this conversion is wrong
bytes(5) creates b"\x00\x00\x00\x00\x00" and not b'\x05\x00\x00\x00'

MartinHjelmare: https://docs.python.org/3/library/stdtypes.html#int.to_bytes

marcelveldt: Yeah exactly but still I wonder if we should do it this way. I just looked up the spec and this should contain the NOC as (bytes encoded) string. So a single int in this field is basically just junk and we should just ignore it. We can parse it to bytes (assuming little endian, unsigned) but its still junk.

marcelveldt: anyways, changed it so we're unblocked.

MartinHjelmare (approved):

tests/common/test_parser.py
marcelveldt: @MartinHjelmare what do you think ? Do the conversion (as at least we get a bytes object back) or just set it to an empty bytes object ?

MartinHjelmare: What does the NOC do?

marcelveldt: > What does the NOC do?

Its a certificate

marcelveldt: So setting it to an empty string feels more safe here to me

marcelveldt: I changed it to only handle this specific edge case - it didn't feel right otherwise as we would be hiding issues.

MartinHjelmare (approved):

PR #686, opened 2024-05-02
Pin Python Matter server dependencies

agners:
Pin the Python Matter server dependencies to specific versions. This avoids surprises. Specifically, also pin orjson to 3.9.15 as newer versions seem to crash on some platforms (observed in the Home Assistant Core project).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,12 +15,12 @@ classifiers = [
   "Topic :: Home Automation",
 ]
 dependencies = [
-  "aiohttp",
-  "aiorun",
-  "async-timeout",
-  "coloredlogs",
-  "dacite",
-  "orjson",
+  "aiohttp==3.9.5",
+  "aiorun==2023.7.2",
+  "async-timeout==4.0.3",
+  "coloredlogs==15.0.1",
+  "dacite==1.8.1",
+  "orjson==3.9.15",
   "home-assistant-chip-clusters==2024.4.1",
 ]
 description = "Python Matter WebSocket Server"


pyproject.toml
```diff
@@ -15,12 +15,12 @@ classifiers = [
   "Topic :: Home Automation",
 ]
 dependencies = [
-  "aiohttp",
-  "aiorun",
-  "async-timeout",
-  "coloredlogs",
-  "dacite",
-  "orjson",
+  "aiohttp==3.9.5",
```
MartinHjelmare, line 18: We don't want to pin any of the client dependencies.

MartinHjelmare (changes requested):

pyproject.toml
agners: Right, to avoid Core getting into conflicting imports. I'll try adding the same dependencies to the server but with constraints.

agners:
I've pinned dependency for the server optional dependencies only, while keeping an "open" dependency for the main package. This seems to work well: When installing just the package, any suitable version will be picked (e.g. by default the latest). When installing the server package, the pinned dependency for the server gets picked.

Of course we can't define conflicting dependencies, pip complains with:
```
The conflict is caused by:
    python-matter-server 0.0.0 depends on orjson==3.9.14
    python-matter-server[server] 0.0.0 depends on orjson==3.9.14
    python-matter-server[server] 0.0.0 depends on orjson==3.9.15; extra == "server"
```

pyproject.toml
```diff
@@ -33,9 +33,15 @@ version = "0.0.0"
 
 [project.optional-dependencies]
 server = [
-  "home-assistant-chip-core==2024.4.1",
+  "aiohttp==3.9.5",
+  "aiorun==2023.7.2",
+  "async-timeout==4.0.3",
+  "coloredlogs==15.0.1",
   "cryptography==42.0.5",
+  "dacite==1.8.1",
```
MartinHjelmare, line 41: Where is dacite used?

agners: Hm, yeah it seems it is no longer used. Dropping it.

MartinHjelmare (approved):
Thanks!

PR #687, opened 2024-05-02
Address and enable flake8 type checking checks

agners:
This mostly moves imports into `if TYPE_CHECKING` blocks which avoids unnecessary imports only used for type checking. These unnecessary imports may affect performance.


MartinHjelmare (approved):
If we think importing the models has an impact, I think this is good. Normally I don't think the imports has that big of an impact.

PR #688, opened 2024-05-02
Deduplicate PAA certificates using Subject Key Identifier

agners:
Matter PAA certificates must have X509 Subject Key Identifier extension. The current code already deduplicated the certificates from the DCL based on this identifier.

This rewrites the code such that certificates fetched from git are deduplicated as well.

### DIFF
--- a/matter_server/server/helpers/paa_certificates.py
+++ b/matter_server/server/helpers/paa_certificates.py
@@ -48,12 +48,12 @@ async def write_paa_root_cert(
         der_certificate: bytes,
     ) -> None:
         # handle PEM certificate file
-        file_path_pem = paa_root_cert_dir.joinpath(filename_base + ".pem")
+        file_path_pem = paa_root_cert_dir.joinpath(f"{filename_base}.pem")
         LOGGER.debug("Writing PEM certificate %s", file_path_pem)
         with open(file_path_pem, "w+", encoding="utf-8") as outfile:
             outfile.write(pem_certificate)
         # handle DER certificate file (converted from PEM)
-        file_path_der = paa_root_cert_dir.joinpath(filename_base + ".der")
+        file_path_der = paa_root_cert_dir.joinpath(f"{filename_base}.der")
         LOGGER.debug("Writing DER certificate %s", file_path_der)
         with open(file_path_der, "wb+") as outfile:
             outfile.write(der_certificate)
--- a/matter_server/server/helpers/paa_certificates.py
+++ b/matter_server/server/helpers/paa_certificates.py
@@ -10,7 +10,6 @@ All rights reserved.
 import asyncio
 from datetime import UTC, datetime, timedelta
 import logging
-import os
 from pathlib import Path
 import re
 import warnings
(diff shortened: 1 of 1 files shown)


agners:
Before this change, there were a total of 442 certificate files:
```
$ ls paa-root-certs-orig/* | wc -l
442
```

But browsing shows there are quite some duplicated. I think most came from the addition of git fetching.

Anyhow, when checking the md5sum, only 272 certificates are actually unique:

```
$  md5sum paa-root-certs-orig/* | sort | cut -d " " -f 1 | uniq | wc -l
272
```

It seems that the Matter stack can handle duplicate certificates just fine, but this seems a bit wasteful.

With this change, the total stored certificates is exactly 272 :tada: :

```
$ ls paa-root-certs/* | wc -l
272
```

agners:
Hm, just realized that the total of downloaded certs does not match the file count:

```
2024-05-02 14:35:42.273 (MainThread) INFO [matter_server.server.helpers.paa_certificates] Fetching the latest PAA root certificates from DCL.
2024-05-02 14:35:44.218 (MainThread) INFO [matter_server.server.helpers.paa_certificates] Fetched 58 PAA root certificates from DCL.
2024-05-02 14:35:51.841 (MainThread) INFO [matter_server.server.helpers.paa_certificates] Fetched 74 PAA root certificates from Test DCL.
2024-05-02 14:35:51.842 (MainThread) INFO [matter_server.server.helpers.paa_certificates] Fetching the latest PAA root certificates from Git.
2024-05-02 14:35:53.290 (MainThread) INFO [matter_server.server.helpers.paa_certificates] Fetched 5 PAA root certificates from Git.
```

This equals to 137 certificates, so this should be stored in 274 files (for pem and der format). So it seems one gets overwritten. Further investigation shows that there is indeed a certificate on the test net with the very same subject, leading to the same name. So the latter overwrites the earlier one :see_no_evil: From looking at the cert, both appear to be valid, just the keys are different. But this is not a new behavior. Also, since it is on the test net, this is likely not all that relevant in practice.

matter_server/server/helpers/paa_certificates.py
```diff
@@ -29,90 +31,117 @@
...
+    def _write(
+        paa_root_cert_dir: Path,
+        filename_base: str,
+        pem_certificate: str,
+        der_certificate: bytes,
+    ) -> None:
         # handle PEM certificate file
-        file_path_pem = f"{filepath_base}.pem"
-        LOGGER.debug("Writing certificate %s", file_path_pem)
+        file_path_pem = paa_root_cert_dir.joinpath(filename_base + ".pem")
```
marcelveldt, line 51:
```suggestion
        file_path_pem = paa_root_cert_dir.joinpath(f"{filepath_base}.pem")
```

matter_server/server/helpers/paa_certificates.py
```diff
@@ -29,90 +31,117 @@
...
+        LOGGER.debug("Writing PEM certificate %s", file_path_pem)
         with open(file_path_pem, "w+", encoding="utf-8") as outfile:
-            outfile.write(certificate)
+            outfile.write(pem_certificate)
         # handle DER certificate file (converted from PEM)
-        pem_certificate = x509.load_pem_x509_certificate(certificate.encode())
-        file_path_der = f"{filepath_base}.der"
-        LOGGER.debug("Writing certificate %s", file_path_der)
+        file_path_der = paa_root_cert_dir.joinpath(filename_base + ".der")
```
marcelveldt, line 56:
```suggestion
        file_path_der = paa_root_cert_dir.joinpath(f"{filepath_base}.der")
```

marcelveldt (approved):
looks good!

matter_server/server/helpers/paa_certificates.py
```diff
@@ -29,90 +31,117 @@
...
+        der_certificate: bytes,
+    ) -> None:
         # handle PEM certificate file
-        file_path_pem = f"{filepath_base}.pem"
-        LOGGER.debug("Writing certificate %s", file_path_pem)
+        file_path_pem = paa_root_cert_dir.joinpath(f"{filename_base}.pem")
+        LOGGER.debug("Writing PEM certificate %s", file_path_pem)
         with open(file_path_pem, "w+", encoding="utf-8") as outfile:
-            outfile.write(certificate)
+            outfile.write(pem_certificate)
```
MartinHjelmare, line 54: Maybe use `Path.write_text`?

matter_server/server/helpers/paa_certificates.py
```diff
@@ -29,90 +31,117 @@
...
         # handle DER certificate file (converted from PEM)
-        pem_certificate = x509.load_pem_x509_certificate(certificate.encode())
-        file_path_der = f"{filepath_base}.der"
-        LOGGER.debug("Writing certificate %s", file_path_der)
+        file_path_der = paa_root_cert_dir.joinpath(f"{filename_base}.der")
+        LOGGER.debug("Writing DER certificate %s", file_path_der)
         with open(file_path_der, "wb+") as outfile:
-            der_certificate = pem_certificate.public_bytes(serialization.Encoding.DER)
             outfile.write(der_certificate)
```
MartinHjelmare, line 59: Maybe use `Path.write_bytes`?

MartinHjelmare (approved):

agners:
> This equals to 137 certificates, so this should be stored in 274 files (for pem and der format). So it seems one gets overwritten. 

For the records: Fixed by appending the Subject Key Identifier to the filename. With that all 274 files are present.

PR #706, opened 2024-05-17
Use mkdirs for paa_root_certs as the intermediate directories do not exist

D3v01dZA:
Fixes https://github.com/home-assistant-libs/python-matter-server/issues/704

### DIFF
--- a/matter_server/server/helpers/paa_certificates.py
+++ b/matter_server/server/helpers/paa_certificates.py
@@ -203,7 +203,7 @@ async def fetch_certificates(
                 if not path.is_dir():
                     path.unlink()
         else:
-            paa_root_cert_dir.mkdir()
+            paa_root_cert_dir.mkdirs()
         return None
 
     last_fetch = await loop.run_in_executor(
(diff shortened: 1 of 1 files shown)


D3v01dZA:
paa_dir is a pathlib.Path

https://github.com/home-assistant-libs/python-matter-server/blob/main/matter_server/server/const.py#L16
https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

marcelveldt (approved):
Looks fine to me, thanks!

petrij98:
Are we able to get this fixed up soon? I'm trying to create a new Python Matter container and this very issue is keeping me from running it.

petrij98:
How are we looking with pushing these changes? I've been trying to get a new Python Matter container running today and have been plagued by this very issue.

olavt:
When will a new build be triggered including this fix?

MadSpindel:
@marcelveldt Can you trigger a new release version with this fix?

marcelveldt:
First of all, not very nice to start tagging people to get them to do a release, especially not on a public holiday.
Second of all, we provide the docker image on a best-effort base, its not officially supported. If you are smart enough that you think you should be running your own docker container to run the Matter server (ignoring all our warnings that running Matter in docker is not as straight forward as you might think) you will be smart enough to just use specific docker tag of the Matter server. Never a good practice to just run "latest" - run a specific version and upgrade with care, especially for docker images that are provided on a best effort base.

PR #709, opened 2024-05-18
Add software update capability

agners:
This adds software update capability to the Python Matter Server.

As it stands, the implementation does the following things:
1. Check the DCL for software updates for a particular device
2. If an update is available, it downloads the firmware locally
3. Prepares the OTA Provider on the Matter fabric (using the Linux OTA Provider example app)
3a. On first use, the OTA Provider App is first commissioned to the fabric
4. Notifies the node about the availability of the OTA Provider and the update

The node then communicates directly with the OTA Provider. Progress of the update can be observed in the OTA Software Update Requestor Cluster of the device being updated.

There are a few things which still need to be sorted out:

- [x] Lifecycle management of the OTA Provider App needs improvement (termination not working/adding more updates)
- [x] Potential candidates need to be better checked if they are indeed compatible (minApplicableSoftwareVersion/maxApplicableSoftwareVersion)
- [x] Checksum of the update files needs to be checked after download
- [x] Split between update check and update apply
- [x] Handle multiple update calls for the same node
- [ ] Support updates from testnet DCL

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN \
 
 ARG PYTHON_MATTER_SERVER
 
-ENV chip_example_url "https://github.com/home-assistant-libs/matter-linux-ota-provider/releases/download/2024.7.0"
+ENV chip_example_url "https://github.com/home-assistant-libs/matter-linux-ota-provider/releases/download/2024.7.1"
 ARG TARGETPLATFORM
 
 RUN \
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN \
     set -x \
     && apt-get update \
     && apt-get install -y --no-install-recommends \
+        curl \
         libuv1 \
         zlib1g \
         libjson-c5 \
@@ -24,6 +25,21 @@ RUN \
         /usr/src/*
 
 ARG PYTHON_MATTER_SERVER
+
+ENV chip_example_url "https://github.com/agners/matter-linux-example-apps/releases/download/v1.3.0.0"
+ARG TARGETPLATFORM
+
+RUN \
+    set -x \
+    && echo "${TARGETPLATFORM}" \
+    && if [ "${TARGETPLATFORM}" = "linux/amd64" ]; then \
+        curl -Lo /usr/local/bin/chip-ota-provider-app "${chip_example_url}/chip-ota-provider-app-x86-64"; \
+    elif [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
+        curl -Lo /usr/local/bin/chip-ota-provider-app "${chip_example_url}/chip-ota-provider-app-aarch64"; \
+    else \
+        exit 1; \
+    fi \
+    && chmod +x /usr/local/bin/chip-ota-provider-app
 
 # hadolint ignore=DL3013
 RUN \
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN \
 
 ARG PYTHON_MATTER_SERVER
 
-ENV chip_example_url "https://github.com/agners/matter-linux-example-apps/releases/download/v1.3.0.0"
+ENV chip_example_url "https://github.com/home-assistant-libs/matter-linux-ota-provider/releases/download/2024.7.0"
 ARG TARGETPLATFORM
 
 RUN \
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -29,6 +29,7 @@ from ..common.models import (
     EventType,
     MatterNodeData,
     MatterNodeEvent,
+    MatterSoftwareVersion,
     MessageType,
     NodePingResult,
     ResultMessageBase,
@@ -509,7 +510,7 @@ class MatterClient:
         """Interview a node."""
         await self.send_command(APICommand.INTERVIEW_NODE, node_id=node_id)
 
-    async def check_node_update(self, node_id: int) -> dict[str, Any]:
+    async def check_node_update(self, node_id: int) -> MatterSoftwareVersion | None:
         """Check Node for updates.
 
         Return a dict with the available update information. Most notable
@@ -518,10 +519,11 @@ class MatterClient:
 
         The "softwareVersionString" is a human friendly version string.
         """
-        node_update = await self.send_command(
-            APICommand.CHECK_NODE_UPDATE, node_id=node_id
-        )
-        return cast(dict[str, Any], node_update)
+        data = await self.send_command(APICommand.CHECK_NODE_UPDATE, node_id=node_id)
+        if data is None:
+            return None
+
+        return dataclass_from_dict(MatterSoftwareVersion, data)
 
     async def update_node(
         self,
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -519,7 +519,7 @@ class MatterClient:
 
         The "softwareVersionString" is a human friendly version string.
         """
-        data = await self.send_command(APICommand.CHECK_NODE_UPDATE, node_id=node_id)
+        data = await self.send_command(APICommand.CHECK_NODE_UPDATE, node_id=node_id, require_schema=10)
         if data is None:
             return None
 
@@ -532,7 +532,10 @@ class MatterClient:
     ) -> None:
         """Start node update to a particular version."""
         await self.send_command(
-            APICommand.UPDATE_NODE, node_id=node_id, software_version=software_version
+            APICommand.UPDATE_NODE,
+            node_id=node_id,
+            software_version=software_version,
+            require_schema=10
         )
 
     def _prepare_message(
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -523,7 +523,11 @@ class MatterClient:
         )
         return cast(dict[str, Any], node_update)
 
-    async def update_node(self, node_id: int, software_version: int) -> None:
+    async def update_node(
+        self,
+        node_id: int,
+        software_version: int | str,
+    ) -> None:
         """Start node update to a particular version."""
         await self.send_command(
             APICommand.UPDATE_NODE, node_id=node_id, software_version=software_version
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -519,7 +519,9 @@ class MatterClient:
 
         The "softwareVersionString" is a human friendly version string.
         """
-        data = await self.send_command(APICommand.CHECK_NODE_UPDATE, node_id=node_id, require_schema=10)
+        data = await self.send_command(
+            APICommand.CHECK_NODE_UPDATE, node_id=node_id, require_schema=10
+        )
         if data is None:
             return None
 
@@ -535,7 +537,7 @@ class MatterClient:
             APICommand.UPDATE_NODE,
             node_id=node_id,
             software_version=software_version,
-            require_schema=10
+            require_schema=10,
         )
 
     def _prepare_message(
--- a/matter_server/client/client.py
+++ b/matter_server/client/client.py
@@ -509,6 +509,26 @@ class MatterClient:
         """Interview a node."""
         await self.send_command(APICommand.INTERVIEW_NODE, node_id=node_id)
 
+    async def check_node_update(self, node_id: int) -> dict[str, Any]:
+        """Check Node for updates.
+
+        Return a dict with the available update information. Most notable
+        "softwareVersion" contains the integer value of the update version which then
+        can be used for the update_node command to trigger the update.
+
+        The "softwareVersionString" is a human friendly version string.
+        """
+        node_update = await self.send_command(
+            APICommand.CHECK_NODE_UPDATE, node_id=node_id
+        )
+        return cast(dict[str, Any], node_update)
+
+    async def update_node(self, node_id: int, software_version: int) -> None:
+        """Start node update to a particular version."""
+        await self.send_command(
+            APICommand.UPDATE_NODE, node_id=node_id, software_version=software_version
+        )
+
     def _prepare_message(
         self,
         command: str,
--- a/matter_server/common/models.py
+++ b/matter_server/common/models.py
@@ -47,6 +47,7 @@ class APICommand(str, Enum):
     PING_NODE = "ping_node"
     GET_NODE_IP_ADDRESSES = "get_node_ip_addresses"
     IMPORT_TEST_NODE = "import_test_node"
+    CHECK_NODE_UPDATE = "check_node_update"
     UPDATE_NODE = "update_node"
 
 
--- a/matter_server/common/models.py
+++ b/matter_server/common/models.py
@@ -47,6 +47,7 @@ class APICommand(str, Enum):
     PING_NODE = "ping_node"
     GET_NODE_IP_ADDRESSES = "get_node_ip_addresses"
     IMPORT_TEST_NODE = "import_test_node"
+    UPDATE_NODE = "update_node"
 
 
 EventCallBackType = Callable[[EventType, Any], None]
--- a/matter_server/common/models.py
+++ b/matter_server/common/models.py
@@ -211,3 +211,21 @@ class CommissioningParameters:
     setup_pin_code: int
     setup_manual_code: str
     setup_qr_code: str
+
+
+@dataclass
+class MatterSoftwareVersion:
+    """Representation of a Matter software version. Return by the check_node_update command.
+
+    This holds Matter software version information similar to what is available from the CSA DCL.
+    https://on.dcl.csa-iot.org/#/Query/ModelVersion.
+    """
+
+    vid: int
+    pid: int
+    software_version: int
+    software_version_string: str
+    firmware_information: str | None
+    min_applicable_software_version: int
+    max_applicable_software_version: int
+    release_notes_url: str | None
--- a/matter_server/common/models.py
+++ b/matter_server/common/models.py
@@ -229,3 +229,30 @@ class MatterSoftwareVersion:
     min_applicable_software_version: int
     max_applicable_software_version: int
     release_notes_url: str | None
+
+    @classmethod
+    def from_dict(cls, data: dict) -> MatterSoftwareVersion:
+        """Initialize from dict."""
+        return cls(
+            vid=data["vid"],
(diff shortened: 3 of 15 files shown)


olavt:
Will there be some new API-commands available to control this?

olavt:
It would be nice if these steps where totally independent:

1. Check the DCL for software updates for a particular device
2. If an update is available, it downloads the firmware locally

I would like them to be:

1. Check the DCL for software updates for a particular device
2. If an update is available go to step 3
3. Download a firmware file from a given location (either one from the one specified by the DCL or any other location)

The reason I want this is to be able to test firmware updates for devices in development, which is not yet published to the DCL. Then step 1 and 2 would be bypassed.

It would be nice with a separate API-function for each of the steps.

marcelveldt:
TODO:

- add separate api command to check for updates
- periodic check for node updates - set flag on matterNodeData
- Show "update available" in Matter UI
- HA update entity attached to the flag + manual service call to check for updates
- The check for updates should implement throttling (HA has helpers for that)

agners:
> Will there be some new API-commands available to control this?

Yes, 

> It would be nice if these steps where totally independent:
> 
>     1. Check the DCL for software updates for a particular device
> 
>     2. If an update is available, it downloads the firmware locally
> 

Yes, I intend to split this two operations. It is a bit a question of what the reference will be to apply the update. I intend to use the software version number (in integer format) as reference, as this should be unique. So the commands will be:

```json
  {
    "command": "check_node_update",
    "args": {
      "node_id": 50
    }
  }
```

This will return information about the software, the exact format is tbd (currently a raw dump of the JSON data returned by the DCL).

Then with this command the update will get applied:

```json
  {
    "command": "update_node",
    "args": {
      "node_id": 50,
      "software_version": 123,
    }
  }
```

Will then start the node update.


> I would like them to be:
> 
>     1. Check the DCL for software updates for a particular device
> 
>     2. If an update is available go to step 3
> 
>     3. Download a firmware file from a given location (either one from the one specified by the DCL or any other location)
> 
> 
> The reason I want this is to be able to test firmware updates for devices in development, which is not yet published to the DCL. Then step 1 and 2 would be bypassed.

Agreed, we'll need to have a way to support non-DCL updates. In fact, for testing I already need this (I've now added hardcoded updates for the OTA Requestor example).

There are various use cases, from just not in DCL but online, to just an update file from somewhere. Each of these cases need some metadata to match the update file with the right devices. The metadata should probably closely follow the DCL format.

For online ones, we could probably just pass it through the WebSocket API as well, essentially how you suggest. But I don't think this will cover all use cases. :thinking:

agners:
image


image

Dockerfile
```diff
@@ -25,6 +26,21 @@ RUN \
 
 ARG PYTHON_MATTER_SERVER
 
+ENV chip_example_url "https://github.com/agners/matter-linux-example-apps/releases/download/v1.3.0.0"
```
marcelveldt, line 29: can we move this to the libs org and at least target it into ota provide only ?

Dockerfile
```diff
@@ -25,6 +26,21 @@ RUN \
...
+    set -x \
+    && echo "${TARGETPLATFORM}" \
+    && if [ "${TARGETPLATFORM}" = "linux/amd64" ]; then \
+        curl -Lo /usr/local/bin/chip-ota-provider-app "${chip_example_url}/chip-ota-provider-app-x86-64"; \
+    elif [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
+        curl -Lo /usr/local/bin/chip-ota-provider-app "${chip_example_url}/chip-ota-provider-app-aarch64"; \
+    else \
+        exit 1; \
+    fi \
+    && chmod +x /usr/local/bin/chip-ota-provider-app
```
marcelveldt, line 42: Do we need to run this at all times in the background ? We can't run it as subprocess from our matter server, basically making the matter server responsible for running it

Dockerfile
agners: > can we move this to the libs org and at least target it into ota provider only ?

Sure.

@frenck can you create a repo in the libs org? I'd name it `matter-linux-ota-provider`.

> Another option: include the final binary for the ota provider in the source code so the source code is at leats tied to a specific version of the ota provider

I am not a big fan in adding binaries into git. Also we'd then have to sort out architecture specifics there.

The Docker approach works nicely IMHO.

Dockerfile
agners: That is not running it, merely setting the executable flag.

We run the OTA provider only during the update. This is handled in `provider.py`.

Dockerfile
marcelveldt: Yeah, figured the fact that we distribute the Matter server as a whole (so the entire docker container/image) this is indeed fine. We should make the readme extra clear that a standalone install is not supported. I'll work a bit more on the docs after I get back from DC.

Dockerfile
marcelveldt: Oof, sorry misread that. I blame jetlag haha. Ignore what I said.

Dockerfile
agners: Yeah if the binary is absent it will lead to an error when trying to update a device. We could preemptively check at startup and add a warning in the logs. But the update entities would still appear :thinking: 

I guess we could make it even a server info (`updates_supported`) and create  update entities on Core side only if the server does support updates. :thinking:

marcelveldt: Well, using the Matter server in HA is not supported from a manual install so this is a non-issue then.

marcelveldt, line 29: did you manage to move it to the libs org ?

matter_server/server/device_controller.py
```diff
@@ -90,11 +98,23 @@
...
     create_attribute_path_from_attribute(
         0, Clusters.BasicInformation.Attributes.SoftwareVersion
     )
 )
+BASIC_INFORMATION_SOFTWARE_VERSION_STRING_ATTRIBUTE_PATH = (
+    create_attribute_path_from_attribute(
+        0, Clusters.BasicInformation.Attributes.SoftwareVersionString
+    )
+)
+
```
marcelveldt, line 117: In a follow-up iteration we should maybe start moving all these constants into the constants module

marcelveldt: well, hmm maybe not as these are not plain/simple constants but created from a util function.

Dockerfile
agners: Moved: https://github.com/home-assistant-libs/matter-linux-ota-provider

matter_server/server/device_controller.py
agners: We probably could add a `to_attribute_path(endpointId: int)` function to the `ClusterAttributeDescriptor`. With this, we could do:

```python
node.attributes.get(str(Clusters.BasicInformation.Attributes.SoftwareVersionString.to_attribute_path(0)))
```

matter_server/client/client.py
```diff
@@ -509,6 +510,31 @@ async def interview_node(self, node_id: int) -> None:
...
+    async def check_node_update(self, node_id: int) -> MatterSoftwareVersion | None:
+        """Check Node for updates.
+
+        Return a dict with the available update information. Most notable
+        "softwareVersion" contains the integer value of the update version which then
+        can be used for the update_node command to trigger the update.
+
+        The "softwareVersionString" is a human friendly version string.
+        """
+        data = await self.send_command(APICommand.CHECK_NODE_UPDATE, node_id=node_id)
```
marcelveldt, line 522:
```suggestion
        data = await self.send_command(APICommand.CHECK_NODE_UPDATE, node_id=node_id, require_schema=10)
```

matter_server/client/client.py
```diff
@@ -509,6 +510,31 @@ async def interview_node(self, node_id: int) -> None:
...
+
+    async def update_node(
+        self,
+        node_id: int,
+        software_version: int | str,
+    ) -> None:
+        """Start node update to a particular version."""
+        await self.send_command(
+            APICommand.UPDATE_NODE, node_id=node_id, software_version=software_version
+        )
```
marcelveldt, line 536:
```suggestion
        await self.send_command(
            APICommand.UPDATE_NODE,
            node_id=node_id,
            software_version=software_version,
            require_schema=10
        )
```

matter_server/common/models.py
```diff
@@ -209,3 +211,21 @@ class CommissioningParameters:
...
+
+
+@dataclass
+class MatterSoftwareVersion:
+    """Representation of a Matter software version. Return by the check_node_update command.
+
+    This holds Matter software version information similar to what is available from the CSA DCL.
+    https://on.dcl.csa-iot.org/#/Query/ModelVersion.
+    """
+
```
marcelveldt, line 223: maybe we can add a source here to distinguish main-net and test-net updates (or other sources maybe)

matter_server/server/ota/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+        "pid": 0x8001,
+        "softwareVersion": 2,
+        "softwareVersionString": "2.0",
+        "cdVersionNumber": 1,
+        "softwareVersionValid": True,
+        "otaChecksum": "7qcyvg2kPmKZaDLIk8C7Vyteqf4DI73x0tFZkmPALCo=",
+        "otaChecksumType": 1,
+        "minApplicableSoftwareVersion": 1,
+        "maxApplicableSoftwareVersion": 1,
+        "otaUrl": "https://github.com/agners/matter-linux-example-apps/releases/download/v1.3.0.0/chip-ota-requestor-app-x86-64.ota",
```
marcelveldt, line 18: want to move this to the libs repo as well to prevent that you need to maintain this on your account ?

matter_server/server/ota/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+        "softwareVersionString": "1.1.11-c85ba1e-dirty",
+        "cdVersionNumber": 1,
+        "softwareVersionValid": True,
+        "otaChecksum": "x2sK9xjVuGff0eefYa4cporDO+Z+WVxxw+JP5Ol+5og=",
+        "otaChecksumType": 1,
+        "minApplicableSoftwareVersion": 10010000,
+        "maxApplicableSoftwareVersion": 10010011,
+        "otaUrl": "https://raw.githubusercontent.com/ChampOnBon/Onvis/master/S4/debug.ota",
+    },
+}
```
marcelveldt, line 34: Question 1: Do you think its a good idea to have this update here as it has a big fat "experimental" vibe around it ;-)

Question 2: So we could extend this list with other hardcoded paths ?

matter_server/server/ota/__init__.py
```diff
@@ -0,0 +1,55 @@
+"""OTA implementation for the Matter Server."""
+
+from matter_server.server.ota import dcl
+
+HARDCODED_UPDATES: dict[tuple[int, int], dict] = {
```
marcelveldt, line 5: another approach could be that we create an ABC for an update provider and then have a base implementation for a "hardcoded update" provider. And that when calling the outer 'check_for_update' we check for updates on all providers. This way you also differentiate between main-net and test-net by making test-net a simple subclass (with a few changed params)

matter_server/server/ota/dcl.py
```diff
@@ -0,0 +1,120 @@
...
+
+from matter_server.common.errors import UpdateCheckError
+from matter_server.server.helpers import DCL_PRODUCTION_URL
+
+LOGGER = logging.getLogger(__name__)
+
+
+async def _get_software_versions(vid: int, pid: int) -> Any:
+    """Check DCL if there are updates available for a particular node."""
+    async with ClientSession(raise_for_status=False) as http_session:
```
marcelveldt, line 17: So, we creating a new http session each time ? Why not use a shared http session at server level ?

matter_server/server/ota/provider.py
```diff
@@ -0,0 +1,366 @@
...
+            )
+            await self._commission_ota_provider(
+                chip_device_controller,
+                ota_provider_passcode,
+                ota_provider_discriminator,
+                ota_provider_node_id,
+            )
+
+            # Notify update node about the availability of the OTA Provider. It will query
+            # the OTA provider and start the update.
```
marcelveldt, line 216:
```suggestion
            # Notify update node about the availability of the OTA Provider. 
            # It will query the OTA provider and start the update.
```

matter_server/server/ota/__init__.py
agners: The examples are also used by the developer add-on so I'd have to change things there too.

We can also delete it here, it is only really useful for testing during development.

matter_server/server/ota/__init__.py
agners: > Question 1: Do you think its a good idea to have this update here as it has a big fat "experimental" vibe around it ;-)

It is what their official app distributes too :man_shrugging:  Just much easier to install compared to their update procedure :see_no_evil: 

FWIW, the firmware seems more reliable, at least my S4 switch stayed online since I've updated.

> 
> Question 2: So we could extend this list with other hardcoded paths ?

Yes we can. But it is hard coded, so I don't think this will scale well. I'd definitely prefer if folks would just use the DCL :see_no_evil:

matter_server/server/ota/dcl.py
agners: Good point! I've moved this up, so we only create one `ClientSession` per update check. But we probably could move it even further up, even make a shared client session for the whole server as you suggest. But if we do that, we should reuse this session in other places too, like vendor information and PAA certificate fetching. I'd rather prefer to do this in a follow up.

matter_server/common/models.py
agners: Yeah good idea, definitely possible. I guess the question will be how do we expose that to the user (as in how can we display it in the update entity). I'd rather prefer to only add if we have actual use for it.

marcelveldt: Well, usecase would be to inform the use in the UI that its a beta update vs stable

matter_server/server/ota/__init__.py
marcelveldt: Whatever you prefer, we can always change it later.

matter_server/server/ota/__init__.py
marcelveldt: Yes, DCL is currently being discussed - te procedures will be simplified soon(ish) but until then we will have to get this going with (temporary) hardcoded lists.

matter_server/server/ota/dcl.py
marcelveldt: Sure, it was just an observeration - write it somewhere down as follow up improvements. 
Not needed for MVP now.

marcelveldt:
Can you mark it as ready once its entirely complete ? Then we'll do one final review (and hopefully the merge)

matter_server/common/models.py
agners: Yes, but *how* would we display that? The Update entity has no such feature currently, from what I know.

I'd imagine update source would be a enum or something, so the question also becomes how we represent extra/additional update sources.

marcelveldt: We don't: 
Its just to steer some custom logic in the integration to show the correct (warning) message for now).
I think what I'm trying to say is that I don't think its of interest to show the exact source to the user in HA (maybe it is in the debug UI) but we can use it to steer other logic.

So the way I see it we have 3 update providers in the code: dcl-main, dcl-test and builtin
Would be good to have that as source attribute on the model (could be an enum as well)

matter_server/server/ota/__init__.py
agners: I've removed the hardcoded updates now and replaced it with capability to load json files from the update directory. This is currently really meant for development. It shouldn't interfere for user updates, which at this point still just supports the main net DCL.

matter_server/common/models.py
agners: Ok, I've added update source as an enum.

marcelveldt (approved):
Looks good!

agners:
> Sorry for chiming in, not sure if this is the best place or if this would be feasible, but: would the updates feature also support firmware downgrading, or that depends on the software on the matter devices?

No, downgrading is not supported by Matter OTA. From the standard:

> OTA Requestors SHALL only upgrade to numerically newer versions and OTA Providers SHALL
only ever provide a numerically newer Software Image than a Node’s current version number (see
SoftwareVersion). Any functional rollback SHALL be affected by the Vendor creating a Software
Image with a newer (higher) version number, but whose binary contents may match prior func­
tionality.

nerijus:
I've built docker image from git main branch and started it - how do I get update notification? It seems there is update available for my device (Aqara Window and Door Sensor P2), about an hour after docker container restart passed, but I don't see any notification in HA.

TheJulianJES:
@nerijus You won't see a notification in HA, as the HA core PR is not merged yet:
- https://github.com/home-assistant/core/pull/120304

nerijus:
Thanks; is there a possibility to upgrade manually from command line from within python-matter-server docker shell?

nerijus:
I updated files from home-assistant/core#120304 manually (and a few other files like select.py, strings.json), but now after restarting HA I get:
```
Error occurred loading flow for integration matter: cannot import name 'UpdateCheckError' from 'matter_server.common.errors' (/usr/local/lib/python3.12/site-packages/matter_server/common/errors.py)
```

marcelveldt:
Just have a little bit of patience folks until this is available. It will also be added as a button the the matter server's own debug UI. The changes for HA core need some more adjustments before we merge them but plan is still to get it done on time for the next beta.

olavt:
I would like to test the new software update capability.

What is the high level flow (which commands to use) if I provide my own firmware file?

JoseAntonioMG:
I'm using HomeAssistant and Matter-Server with thread devices, everything works fine. I would like to be able to perform OTA updates to the devices but I don't have an account in DCL, is there a possibility to update the firmware of the devices added in HomeAssistant through Matter-Server locally? for example by indicating to Matter-Server a path where the update files are. Is there such a possibility?

I have seen that in the Matter-Server configuration there is an option to add "--ota-provider-dir", but I don't know how it works. Is this option the correct one to indicate a path where the update files are saved? How do you indicate that path?

thanks

hoppel118:
@JoseAntonioMG Look here:

https://github.com/home-assistant-libs/python-matter-server/issues/833

PR #715, opened 2024-05-21
Refactor logic for handling Custom Clusters/Attributes

marcelveldt:
- Refactor read_attribute so it can take both single attribute path and multiple (including wildcards)
- Move custom clusters into shared/common logic, accessible by both server and client
- Add logic to poll (required) custom clusters from the server
- Rename Eve custom cluster as its a Vendor-wide cluster, not only for Eve Energy

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -605,33 +605,58 @@ class MatterDeviceController:
 
     @api_command(APICommand.READ_ATTRIBUTE)
     async def read_attribute(
-        self, node_id: int, attribute_path: str, fabric_filtered: bool = False
+        self,
+        node_id: int,
+        attribute_path: str | list[str],
+        fabric_filtered: bool = False,
     ) -> dict[str, Any]:
-        """Read one or more attribute(s) on a node by specifying an attributepath."""
+        """
+        Read one or more attribute(s) on a node by specifying an attributepath.
+
+        The attribute path can be a single string or a list of strings.
+        The attribute path may contain wildcards (*) for cluster and/or attribute id.
+
+        The return type is a dictionary with the attribute path as key and the value as value.
+        """
         if self.chip_controller is None:
             raise RuntimeError("Device Controller not initialized.")
         if (node := self._nodes.get(node_id)) is None or not node.available:
             raise NodeNotReady(f"Node {node_id} is not (yet) available.")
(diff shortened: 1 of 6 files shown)


matter_server/server/device_controller.py
```diff
@@ -1256,6 +1275,13 @@ def log_node_long_setup(time_start: float) -> None:
                     )
                     # NOTE: the node will be picked up by mdns discovery automatically
                     # when it becomes available again.
+                    return
+
+                # check if this node has any custom clusters that need to be polled
+                if polled_attributes := check_polled_attributes(node_data):
+                    self._polled_attributes[node_id] = polled_attributes
```
agners, line 1282: Should we remove the node from `self._polled_attributes` when the node goes unavailable?

matter_server/server/device_controller.py
```diff
@@ -1500,3 +1526,55 @@ def _get_node_lock(self, node_id: int) -> asyncio.Lock:
...
+                        attribute_path,
+                        node_id,
+                        str(err) or err.__class__.__name__,
+                        # log full stack trace if verbose logging is enabled
+                        exc_info=err
+                        if LOGGER.isEnabledFor(VERBOSE_LOG_LEVEL)
+                        else None,
+                    )
+                # polling attributes is heavy on network traffic, so we throttle it a bit
+                await asyncio.sleep(5)
```
agners, line 1555: This seems rather conservative to me, given that we poll 3 or 4 almost concurrently currently. Maybe 2s?

agners (commented):
It is currently useful for custom energy attributes, but in general, I really hope that we don't need this attribute poller a lot :see_no_evil:  :crossed_fingers:

matter_server/server/device_controller.py
marcelveldt: Yeah, I wasn't sure about good/sane values. This is maybe a step too far :-)
I'll change it to 2s indeed.

matter_server/server/device_controller.py
marcelveldt: It should skip unavailable nodes already and if we remove it we need to call the expensive function again to figure out attributes to poll

agners: Ah I see `_custom_attributes_poller` is using `continue` when not available. Yeah make sense :+1:

agners (approved):

PR #716, opened 2024-05-23
Use SetRawAttributeUpdateCallback to support custom clusters

agners:
Update the CHIP wheels to 2024.5.1 to get the new
RawAttributeUpdateCallback.

To support subscriptions for custom cluster use this new RawAttributeUpdateCallback instead of AttributeUpdateCallback. This also makes sure that we update the attributes with the same objects on initial read and subsequent subscription updates (Python types but not Cluster specific struct types etc.).

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -88,6 +88,14 @@ TEST_NODE_START = 900000
 ROUTING_ROLE_ATTRIBUTE_PATH = create_attribute_path_from_attribute(
     0, Clusters.ThreadNetworkDiagnostics.Attributes.RoutingRole
 )
+DESCRIPTOR_PARTS_LIST_ATTRIBUTE_PATH = create_attribute_path_from_attribute(
+    0, Clusters.Descriptor.Attributes.PartsList
+)
+BASIC_INFORMATION_SOFTWARE_VERSION_ATTRIBUTE_PATH = (
+    create_attribute_path_from_attribute(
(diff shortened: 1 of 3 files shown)


matter_server/server/device_controller.py
```diff
@@ -1116,7 +1118,14 @@ def resubscription_succeeded(
...
             )
 
-        sub.SetAttributeUpdateCallback(attribute_updated_callback)
+        if not isinstance(sub, Attribute.SubscriptionTransaction):
+            # Aborted setups result in ReadResult instead of SubscriptionTransaction
+            # Probably a bug: https://github.com/project-chip/connectedhomeip/issues/33570
+            node_logger.warning("Subscription setup failed.")
+            return
+
+        sub.SetAttributeUpdateCallback(None)
```
marcelveldt, line 1127: this feels a bit weird. Why is the callback explicitly set to None here ? Isn't it None/undefined by default ?

agners: Exactly, there is a default handler configured currently, which prints changes to stdout :see_no_evil: 

I could have patched that out in our CHIP SDK changes, but I wanted to keep changes minimal for now.

marcelveldt: can you leave a comment above this line then as a reminder for later ?

agners, line 1127:
```suggestion
        # Make sure to clear default handler which prints to stdout
        sub.SetAttributeUpdateCallback(None)
```

PR #717, opened 2024-05-24
Refactor and rename _resolve_node to _establish_case_session

agners:
The function _resolve_node calls GetConnectedDevice(), which effectively uses the CASE (Certificate Authenticated Session Establishment) protocol to establish a session with the device. Rename the function accordingly.

While at it, convert it to a loop and print time establishment took with higher precision (with one decimal place).

### DIFF
--- a/matter_server/common/errors.py
+++ b/matter_server/common/errors.py
@@ -42,7 +42,7 @@ class NodeNotReady(MatterError):
 
 
 class NodeNotResolving(MatterError):
-    """Error raised when resolving the node fails."""
+    """Error raised when no CASE session could be established."""
 
     error_code = 4
 
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -671,7 +671,7 @@ class MatterDeviceController:
         # This is basically a re-implementation of the chip controller's Read function
         # but one that allows us to send/request custom attributes.
         future = self.server.loop.create_future()
-        device = await self._resolve_node(node_id)
+        device = await self._establish_case_session(node_id)
         async with self._get_node_lock(node_id):
             Attribute.Read(
                 future=future,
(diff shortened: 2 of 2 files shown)


marcelveldt:
I'm not sure about this, sorry.

This step does two things: resolve the node and if that succeeded, establish the case session.
If there is already an existing case session, it will return early with the handle to the existing session.

In any way, "resolve node" feels like a better describing name (also for outsiders) then "establish case session".
Also, not sure what benefit the loop brings in terms of readability.

Compromis: rename the function to "rename node" and explain in its docstring exactly what it does.

marcelveldt (changes requested):
See comment - I dont think we should rename it to this.

agners:
> This step does two things: resolve the node and if that succeeded, establish the case session.

Well, not exactly: It checks the CASE session manager for an existing session, and returns it if one exists. If none exists, it tries to establish one. In other words, it doesn't always resolves the node. If there is an existing session, even though the node is offline, it just returns that (stale) session.

That is exactly the reason I want to rename this function: I want to avoid that the user makes wrong assumptions.

> In any way, "resolve node" feels like a better describing name (also for outsiders) then "establish case session".
Also, not sure what benefit the loop brings in terms of readability.

When you have a function which downloads a file from a URL, the first thing the http library will be doing is resolve the hostname to an IP, then connecting to that host to download the file. Such a function should be named `download_file_from_url()` or something.

A function should always be named what it is doing on a overarching/high level.


That being said, `_establish_case_session()` is maybe also not super accurate. The underlying SDK function is named `mSystemState->CASESessionMgr()->FindOrEstablishSession()`, maybe `_find_or_establish_case_session()` is most appropriate. :thinking:

agners:
Btw, we call `GetConnectedDevice()` from the Python library, but it seems to be rather generic name without much information, so I rather prefer to not follow that naming. :cry:

marcelveldt:
I like FindOrEstablishSession, especially if that is what its called in the SDK.
GetConnectedDevice is indeed not good.

marcelveldt (approved):

PR #720, opened 2024-05-28
Refactor MatterDeviceController

agners:
This commit renames the current MatterDeviceController to MatterDeviceControllerAPI as large parts of the code deal with the WebSocket API. A new class MatterDeviceController got introduced which abstracts the ChipDeviceController in a nicer way. It encapsulates the async friendly call to the SDK (in an executor for those API where this is still necessary). It also takes care of locking per node and manages the subscriptions.

To MatterDeviceControllerAPI it exposes a more Pythonic API to make our code more coherent.


matter_server/server/server.py
```diff
@@ -123,7 +128,10 @@ def __init__(
         self.stack = MatterStack(self)
         # Initialize our (intermediate) device controller which keeps track
         # of Matter devices and their subscriptions.
-        self.device_controller = MatterDeviceController(self)
+        self.device_controller = MatterDeviceController(self, self.paa_root_cert_dir)
+        self.device_controller_api = MatterDeviceControllerAPI(
+            self, self.device_controller
```
marcelveldt, line 133: Now we're refactoring anyways, I would make these private and expose them publically using getters (properties)

tests/server/test_server.py
```diff
@@ -102,7 +102,7 @@ def storage_controller_fixture() -> Generator[MagicMock, None, None]:
 def fetch_certificates_fixture() -> Generator[MagicMock, None, None]:
     """Return a mocked fetch certificates."""
     with patch(
-        "matter_server.server.device_controller.fetch_certificates", autospec=True
+        "matter_server.server.server.fetch_certificates", autospec=True
```
marcelveldt, line 105: ?

matter_server/server/device_controller_api.py
```diff
@@ -0,0 +1,1413 @@
+"""API for controller that Manages Matter devices."""
```
marcelveldt, line 1: Can we make this docstring a bit more clear to explain to first-viewers what thius does and more important whats the diference with the device controller ?

matter_server/server/device_controller.py
```diff
@@ -1,613 +1,280 @@
...
-    )
-)
-
-
-# pylint: disable=too-many-lines,too-many-locals,too-many-statements,too-many-branches,too-many-instance-attributes
 
 
 class MatterDeviceController:
-    """Class that manages the Matter devices."""
+    """Class exposes a Matter devices controller features.
```
marcelveldt, line 39:
```suggestion
    """Class exposing Matter devices controller features.
```

tests/server/test_server.py
agners: In theory, we could run two Device controller (mutiple fabrics). But even in that case, we would only want to fetch once. So I've moved the PAA certificate fetching to the server. I think it makes more sense there, as this is a "server wide" problem. No need to clutter the device controller with it.

matter_server/server/device_controller_api.py
agners: Done :+1:

tests/server/test_server.py
marcelveldt: yeah, but there is two time server now - shouldnt that be just matter_server.server.fetch_certificates

marcelveldt: ah no, ignore me this is module level ofc

marcelveldt (approved):
Super nice!
Thanks for the patience having the back and forth discussion about the design.

This is definitely a good step in making the design more maintainable, well done.

marcelveldt (approved):

PR #723, opened 2024-05-29
Use GetConnectedDevice directly to get device proxy

agners:
Instead of using find_or_establish_case_session use the GetConnectedDevice API directly to get a device proxy for attribute reads. This has the advantage that it runs in a single node lock. With that we no longer will try to establish a case session multiple times though.

### DIFF
--- a/matter_server/server/sdk.py
+++ b/matter_server/server/sdk.py
@@ -319,8 +319,16 @@ class ChipDeviceControllerWrapper:
         # This is basically a re-implementation of the chip controller's Read function
         # but one that allows us to send/request custom attributes.
         future = self.server.loop.create_future()
-        device = await self.find_or_establish_case_session(node_id)
         async with self._get_node_lock(node_id):
+            # GetConnectedDevice is guaranteed to return a deviceProxy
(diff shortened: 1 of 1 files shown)


marcelveldt:
@agners what's the status of this PR, can it move forward ?

agners:
> @agners what's the status of this PR, can it move forward ?

The Matter SDK allows to set `attemptCount` with `FindOrEstablishSession`. Ideally I would like to add that first. then we even have the retry mechanism in all cases where a session is required.

agners:
> In the read this is fine as then we are already (pretty much) certain there should be a handle to the device.
For commissioning and node_setup its not as there we can run into race conditions with resolving (so then we need the retries).

Yeah I think that is a good point, in read there is really less incentives to do a retry. The read is currently mostly (only?) used by the attribute polling, which will poll again anyways. We also have the subscription alive for all devices. So the device will eventually get offline and then the node_setup would kick in, which still has the retry logic.

So I think this is good to go as is :+1:

marcelveldt (approved):

PR #728, opened 2024-05-29
Fix SDK reentrancy issues

agners:
API calls to the native side of the SDK which have global callback functions called from the native side in ChipStack are not re-entrant since they use a shared Future object to signal the result. Currently these are all Commissioning and the Open Commission Window APIs. They can also be identified if they call CallAsyncWithCompleteCallback internally. Make sure those are called from a single Thread only.

Also bump the SDK which contains a fix where the same Future is unnecessarily cleared/set for regular API calls. This avoid re-entrancy issues between regular API calls and the ones with callbacks from the native side.

With this, all currently known re-entrancy issues should be addressed.

### DIFF
--- a/matter_server/server/sdk.py
+++ b/matter_server/server/sdk.py
@@ -59,7 +59,7 @@ class ChipDeviceControllerWrapper:
 
         self._node_lock: dict[int, asyncio.Lock] = {}
         self._subscriptions: dict[int, Attribute.SubscriptionTransaction] = {}
-        self._sdk_non_entrant_executer = ThreadPoolExecutor(max_workers=1)
+        self._sdk_non_entrant_executor = ThreadPoolExecutor(max_workers=1)
 
         # Instantiate the underlying ChipDeviceController instance on the Fabric
         self._chip_controller = self.server.stack.fabric_admin.NewController(
@@ -107,7 +107,7 @@ class ChipDeviceControllerWrapper:
         **kwargs: Any,
     ) -> _T:
         return await self._call_sdk_executor(
-            self._sdk_non_entrant_executer, target, *args, **kwargs
+            self._sdk_non_entrant_executor, target, *args, **kwargs
         )
 
     async def get_compressed_fabric_id(self) -> int:
--- a/matter_server/server/sdk.py
+++ b/matter_server/server/sdk.py
@@ -8,6 +8,7 @@ also makes the API more pythonic where possible.
 from __future__ import annotations
 
 import asyncio
+from concurrent.futures import ThreadPoolExecutor
 from functools import partial
 import logging
 import time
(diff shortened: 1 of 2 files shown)


balloob (approved):

matter_server/server/sdk.py
```diff
@@ -58,6 +59,7 @@ def __init__(self, server: MatterServer, paa_root_cert_dir: Path):
 
         self._node_lock: dict[int, asyncio.Lock] = {}
         self._subscriptions: dict[int, Attribute.SubscriptionTransaction] = {}
+        self._sdk_non_entrant_executer = ThreadPoolExecutor(max_workers=1)
```
marcelveldt, line 62:
```suggestion
        self._sdk_non_entrant_executor = ThreadPoolExecutor(max_workers=1)
```

matter_server/server/sdk.py
```diff
@@ -168,7 +189,7 @@ async def open_commissioning_window(
     ) -> CommissioningParameters:
         """Open a commissioning window to commission a device present on this controller to another."""
         async with self._get_node_lock(node_id):
-            return await self._call_sdk(
+            return await self._call_sdk_non_reentrant(
```
marcelveldt, line 192: I was wondering if these are all cases where we need to ensure the same thread accesses the underlying sdk handle. If we are not entirely sure that the other cases are safe, why not just play it safe and always use this single-thread/instance executor all the time ? We have moved a lot of calls already to their async equivalents anyways already.

agners: Well, that is what I did in #599, but you've reverted it in #621, on the grounds of:

> Now that we have limited the executor to a single executor, we're blocking concurrent actions.
For example if we're busy resolving a node, we cant execute any device commands at the same time which gives a bad UX.

In the end it doesn't matter much, as there are only a few, afaik very short running calls left (`SetWiFiCredentials()`, `SetThreadOperationalDataset()`, `GetAddressAndPort()`, and some `Shutdown()` calls).

If we move all calls to the SDK to a single threaded executor, these can be (IMHO unnecessarily) be blocked by an ongoing Commissioning attempt.

Even when using the single-threaded executor for all SDK calls, we still call from two different Threads: the main asyncio event loop and the executor. Both use `Call()` underneath, which, if not re-entrant, could lead to the very same problem still :man_shrugging: 

In the end, this is all just a stop-gap solution: I want to convert all API's we use to proper co-routines. With that we will make call calls to the SDK from one Thread only (the asyncio event loop) avoid unnecessary Thread switches and side-step any re-entrancy issues.

agners: In the end, either way works for me. But I think this is the change with the least amount of possible surprises, while still fixing all re-entrance issues I am aware of. :sweat_smile:

marcelveldt: Yeah, I mean for all ongoing operations we already switched to the async variants already right ? I was just thinking that if we are not 100% certain about thread safety, we should play it safe. Also, not only threads but asyncio tasks can access objects from multiple tasks at once so it remains a tricky one

marcelveldt (approved):

PR #733, opened 2024-06-03
Don't fail when encountering invalid strings

agners:
Technically all character strings are supposed to be valid UTF-8 strings. We've encountered devices which return non-printable characters as part of the octet string. This change makes the client more robust and not fail when encountering such strings.


frenck (approved):

agners:
In particular, this has been observed on early tado Thermostat X devices (firmware 181.2).

PR #734, opened 2024-06-03
Improve Thread network diagnostics parsing

agners:
Not all Thread devices seem to provide the Thread Network Diagnostics Cluster. According to the Matter specification, this cluster is optional, even for Thread devices.

This commit only parses Thread Network Diagnostics data if the cluster is present in the device.

For completeness, also make sure that Router Eligible End Device (REED) are classified as end devices too (since that is their current role).


agners:
Similar to #733, this has been observed by tado Thermostat X devices with (firmware 219.1).

frenck (approved):

PR #735, opened 2024-06-05
Fix and improve node unavailable handling

agners:
Besides the existing cases, also listen for the BasicInformation Shutdown event. This is usually called when a node gracefully is being shutdown, e.g. after an update.

During testing, I've noticed that the Subscription callback closures capture the MatterNodeData object  in `_subscribe_node()`. This causes the callbacks to potentially update a stale version of the MatterNodeData objects, e.g. when trigger a re-interview: A re-interview doesn't cause the subscriptions to update, hence they still operate on the stale MatterNodeData object. Make sure to always get the MatterNodeData object in each callback closure.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -877,7 +877,6 @@ class MatterDeviceController:
             )
 
         node_logger = LOGGER.getChild(f"node_{node_id}")
-        node = self._nodes[node_id]
 
         # Shutdown existing subscriptions for this node first
         await self._chip_device_controller.shutdown_subscription(node_id)
@@ -896,6 +895,7 @@ class MatterDeviceController:
             )
 
             # work out added/removed endpoints on bridges
+            node = self._nodes[node_id]
             if node.is_bridge and str(path) == DESCRIPTOR_PARTS_LIST_ATTRIBUTE_PATH:
                 endpoints_removed = set(old_value or []) - set(new_value)
                 endpoints_added = set(new_value) - set(old_value or [])
@@ -939,6 +939,7 @@ class MatterDeviceController:
             if isinstance(new_value, ValueDecodeFailure):
                 return
 
+            node = self._nodes[node_id]
             old_value = node.attributes.get(str(path))
 
             # return early if the value did not actually change at all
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -969,6 +970,10 @@ def event_callback(
                 data=data.Data,
             )
             self.event_history.append(node_event)
+
+            if isinstance(data.Data, Clusters.BasicInformation.Events.ShutDown):
+                loop.call_soon_threadsafe(self._node_unavailable, node_id)
```
balloob, line 975: can this be grouped in next threadsafe call ?

agners: Since we don't return in any of this code, we'll always call back to the main Thread. I've moved all processing to the main thread, this makes it much easier to deal with more changes to event processing in the future.

matter_server/server/device_controller.py
```diff
@@ -949,11 +950,10 @@ def attribute_updated_callback(
                 attribute_updated, path, old_value, new_value
             )
 
-        def event_callback(
+        def event(
```
marcelveldt, line 953: Where is this called ?

marcelveldt: Looking at the code this seems to be dead code as the logic is handled in "event_callback"

marcelveldt (changes requested):
see inline comment about the event callback

matter_server/server/device_controller.py
agners: This is called via `self._loop.call_soon_threadsafe`. The diff is a bit hard to read, but essentially everything in `event_callback` was moved to the function just named `event`.

marcelveldt: OK, in that case maybe the function needs to be renamed to "on_event" or something like that to indicate its still a callback ?

agners: Agreed, just `event` is really confusing. But it did align well with what we had with `attribute_updated_callback` (and `attribute_updated`).

Essentially, what we do here is we convert a threaded callback into a callback on the event loop. So I came up with this now:

- `attribute_updated_callback_threadsafe` calls `attribute_updated_callback` on the event loop
- `event_callback_threadsafe` calls `event_callback` on the event loop

agners: Btw, I did check the code, these are the only two callbacks (from the subscription callbacks) which are called straight from the CHIP SDK thread currently.

balloob: Anything called on the event loop should be prefixed with `async_`

agners: Hm, I'd say that should essentially be pretty much everything in this case no :thinking: . Does that make sense for a project which started off with asyncio?

marcelveldt: > Anything called on the event loop should be prefixed with `async_`

In HA core I agree (as there is still a large part non-async) but our code base is started with asyncio in mind and we are talking about a handful of non-async calls so if imo prefixing with async_ is a bit meh. I'd rather prefix the non-async friendly ones or mention in the docstring that its not async friendly.

marcelveldt (approved):

matter_server/server/device_controller.py
agners: > Anything called on the event loop should be prefixed with `async_`

Ok, we kinda went the other way here, and essentially marked thread safe functions now.

I think it is already better what we have, so I'll move forward with this as is. IMHO, if we decide to prefix everything with `async_` would be a bigger change and should be a separate PR.

PR #736, opened 2024-06-06
Use secrets to generate the discriminator

agners:
Use the secrets module which is cryptographically secure to generate the discriminator. The discriminator probably doesn't need to be cryptographically safe, but it also doesn't hurt and gets rid of a QA exception.


agners:
I am using the secrets module also in #709, where I need to generate a passcode. The passcode should be cryptographically safe, so I think it is better to use the same module everywhere.

balloob (approved):

PR #737, opened 2024-06-06
Introduce loop property to get the event loop

agners:
Instead of checking self.server.loop in various places introduce a property to get the event loop.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -225,10 +225,9 @@ class MatterDeviceController:
         return self._thread_credentials_set
 
     @cached_property
-    def loop(self) -> asyncio.AbstractEventLoop:
+    def _loop(self) -> asyncio.AbstractEventLoop:
         """Return the event loop."""
-        if TYPE_CHECKING:
-            assert self.server.loop
+        assert self.server.loop
         return self.server.loop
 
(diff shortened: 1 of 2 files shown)


matter_server/server/device_controller.py
```diff
@@ -223,6 +223,13 @@ def thread_credentials_set(self) -> bool:
         """Return if Thread operational dataset as been set."""
         return self._thread_credentials_set
 
+    @property
```
balloob, line 226: it never changes, so you could do `@cached_property`. Or, go one step further and just set `self.loop = …` in constructor.

balloob (approved):

matter_server/server/device_controller.py
agners: > Or, go one step further and just set self.loop = … in constructor.

That is not possible as the loop on the server is not set at that time.

We do create all objects in constructors, at which time the event loop isn't started yet.

I go for the cached property for now :+1:

agners: `@cached_property` is a bit problematic with our API registration function: It used to access the `loop` property before loop as set on server side. I've marked it private (with `_`) and make sure that any private method is skipped (single `_` instead of `__`).

balloob (approved):

PR #738, opened 2024-06-07
Update to Matter SDK wheels 2024.6.0

agners:
Matter SDK/CHIP wheels [2024.6.0 release notes](https://github.com/home-assistant-libs/chip-wheels/releases/tag/2024.6.0).

This mostly aligns the Matter SDK Python bindings to what is in latest master.

Probably most noteworthy is that SendCommand/GetConnectedDevice are now asyncio friendly throughout, so should never block even when the CHIP thread is very busy. No such cases have been observed, so this is mostly theoretical at this point.

Some unnecessary locking has been dropped which potentially could improve performance.

And finally, the API `TriggerResubscribeIfScheduled` on subscriptions is exposed to Python. This allows to resubscribe subscriptions quicker when a device re-appears.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1301,6 +1301,15 @@ class MatterDeviceController:
         elif (now - last_seen) > NODE_MDNS_BACKOFF:
             # node came back online after being offline for a while or restarted
             logger.info("Node %s re-discovered on MDNS", node_id)
+        elif state_change == ServiceStateChange.Added:
+            # Trigger node re-subscriptions when mDNS entry got added
+            logger.info("Node %s activity on MDNS, trigger resubscribe", node_id)
+            asyncio.create_task(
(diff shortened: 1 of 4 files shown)


frenck (approved):

matter_server/server/device_controller.py
```diff
@@ -1300,6 +1300,15 @@ def _on_mdns_operational_node_state(
         elif (now - last_seen) > NODE_MDNS_BACKOFF:
             # node came back online after being offline for a while or restarted
             logger.info("Node %s re-discovered on MDNS", node_id)
+        elif state_change == ServiceStateChange.Added:
+            # Trigger node re-subscriptions when mDNS entry got added
+            logger.info("Node %s activity on MDNS, trigger resubscribe", node_id)
+            asyncio.create_task(
+                self._chip_device_controller.trigger_resubscribe_if_scheduled(
```
marcelveldt, line 1307: This is nice !

PR #745, opened 2024-06-10
Bump schema version to 9

agners:
With #715, the schema actually changed as an API (subscribe_attribute) got removed. Retroactively bump the schema version to 9.

### DIFF
--- a/matter_server/server/const.py
+++ b/matter_server/server/const.py
@@ -4,7 +4,7 @@ import pathlib
 from typing import Final
 
 # The minimum schema version (of a client) the server can support
-MIN_SCHEMA_VERSION = 5
+MIN_SCHEMA_VERSION = 9
 
 # schema version of our data model
 # only bump if the format of the data in MatterNodeData changed


matter_server/server/const.py
```diff
@@ -4,7 +4,7 @@
 from typing import Final
 
 # The minimum schema version (of a client) the server can support
-MIN_SCHEMA_VERSION = 5
+MIN_SCHEMA_VERSION = 9
```
marcelveldt, line 7: This defines what client version we can minimally support - the actual schema version is defined in the common constants

marcelveldt (changes requested):

matter_server/server/const.py
agners: Yes, and #715 already did bump the schema version to 9, see https://github.com/home-assistant-libs/python-matter-server/pull/715/files#diff-cc3011487c100cb43b984af34c48a3b16cf6ba7fd0f11274278b1502dcaa8b80

This change tells a client which is older (e.g. 8) that we can no longer support that particular client (since #715 removed the `subscribe_attribute` API command).

agners: Hm, I see, my PR title/description was really unclear. I've updated the PR title/description.

marcelveldt: Yeah, ll clear now - the title tricked me indeed.

PR #752, opened 2024-06-13
Improve the open commissioning window command

agners:
Check if the commissioning window is indeed open before returning cached values. This is to prevent returning stale values when the commissioning window got closed, either because the device got successfully commissioned with another network or because the commissioning window got closed due to an error in attempting to commission. Especially the latter case leads to very bad user experience as folks try to use the stale (non-working) commissioning code for the next 10 minutes and get frustrated.

While at it, also close the commissioning window if someone else opened it and we can't provide credentials.


marcelveldt (approved):
Nice!

PR #755, opened 2024-06-14
[Eve Door] EveCluster --> Attribute TimesOpened

lboue:
**EveCluster**

I would like to add attributes to the "EveCluster" cluster. You can see them here:

EveCluster Attributes for Eve Door:
- TimesOpened: 319422470 (0x00130a0006)

image


marcelveldt (approved):
Thanks @lboue !

PR #759, opened 2024-06-17
Do not filter out subscription updates by fabric

marcelveldt:
While testing a Matter smart lock (Aqara U200) I noticed that state changes do not propagate when the lock is controlled from the keypad or another controller. Turns out that the state changes are not coming through as attribute update if the command was not issued by our controller, due to the fact we filter attribute updates by default by fabric.

This PR updates that behavior so we will also be notified by state changes originated from other controllers.
Other than maybe a little bit of extra traffic I don't see any side effects from this change.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1094,6 +1094,7 @@ class MatterDeviceController:
                 return_cluster_objects=False,
                 report_interval=(interval_floor, interval_ceiling),
                 auto_resubscribe=True,
+                fabric_filtered=False,
             )
         )
 


agners:
> Turns out that the state changes are not coming through as attribute update if the command was not issued by our controller, due to the fact we filter attribute updates by default by fabric.

I guess that is `Clusters.DoorLock.Attributes.LockState`? 

I've just tried that with the SiLabs lock example app: Commissioned with two fabrics, locked from one fabric and I do seem to get the state update on the other fabric:

```
2024-06-17 14:23:43.882 (MainThread) VERBOSE [matter_server.server.device_controller.node_30] Attribute updated: 1/257/0 - old value: 2 - new value: 1
```

Which makes me wonder: Is one of the lock not standard conformant? What lock is correct here? :thinking: 

> Other than maybe a little bit of extra traffic I don't see any side effects from this change.

I guess that is the big question, how much (potential) more traffic can this be?

marcelveldt:
Its worth a question in the Matter TT perhaps ?

marcelveldt:
Lets hold off on this one and investigate the usage of the node event instead

marcelveldt:
This one can be closed for now (until we hit the next edge case)

PR #761, opened 2024-06-17
Use UnpairDevice() Matter SDK API on device removal

agners:
Instead of implement our own fabric removal logic, use the existing UnpairDevice() API from Matter SDK. The implementation currently does the same as we did, but it is more robust and future-proof.


marcelveldt:
Ah! That didnt exist back then when we implemented this, nice!

marcelveldt (approved):
LGTM, nice!

PR #762, opened 2024-06-17
Added custom cluster for energy consumption data for Neo smart plug

jf-wm:
I added a custom cluster definition for energy consumption data that the Neo Coolcam Matter smart plug measures and writes to this cluster.

However the raw values for voltage and watt will need additional scaling, the voltage raw value for e.g. 227.3 V is "2273", the watt raw value for e.g. 10.1 W is "101". Current is reported in mA. Wattaccumulated is reported in Wh, however the time frame for each accumulation measurement does not seem to be constant.  

In addition I could not yet find a raw value for total energy consumption, even so the Tuya app reports such a value.

### DIFF
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -237,11 +237,12 @@ class EveCluster(Cluster, CustomClusterMixin):
 
             value: float32 = 0
 
+
 @dataclass
 class NeoCluster(Cluster, CustomClusterMixin):
     """Custom (vendor-specific) cluster for Neo - Vendor ID 4991 (0x137F)."""
 
-    id: ClassVar[int] = 0x00125dfc11
+    id: ClassVar[int] = 0x00125DFC11
 
     @ChipUtility.classproperty
     def descriptor(cls) -> ClusterObjectDescriptor:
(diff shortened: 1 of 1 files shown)


marcelveldt:
The scaling can be done in the Home Assistant discovery schema (once we have bumped the lib there with a version that includes this change)

matter_server/common/custom_clusters.py
```diff
@@ -237,6 +237,132 @@ def attribute_type(cls) -> ClusterObjectFieldDescriptor:
...
+    current: float32 | None = None
+
+    class Attributes:
+        """Attributes for the Neo Cluster."""
+
+        @dataclass
+        class Watt(ClusterAttributeDescriptor, CustomClusterAttributeMixin):
+            """Watt Attribute within the Neo Cluster."""
+
+            should_poll = True
```
marcelveldt, line 278: Does it really need to be polled ? The device is not emitting an attribute update event if the value changes ?

jf-wm: If the value changes significantly the device will emit an update almost immediately, however when the value only changes marginally (within a few W) updates will not always be emitted. Therefore I thought it would be useful to also poll this value.

marcelveldt: OK, well, can you then leave it out for the other attributes maybe ? As polling comes with a bit of overhead so only do it if the device doesn't proactively report it at all (such as Eve)

jf-wm: I would suggest that we leave it out for all attributes initialy to limit overhead. If the data is disturbed afterwards because the device emites to infrequently for low value changes, i will commit a change request to set polling to true.

marcelveldt: sounds good, thanks!

marcelveldt: I've edited the PR to leave out the should_poll, this is all set now.

marcelveldt (approved):
Thanks @jf-wm !

marcelveldt:
Don't forget that we will also need to add this to the HA discovery schema, once the library version has been bumped.
Check Eve for an example: https://github.com/home-assistant/core/blob/dev/homeassistant/components/matter/sensor.py#L173

marcelveldt:
@jf-wm FYI: https://github.com/home-assistant/core/pull/121080/files

PR #765, opened 2024-06-18
Convert to asyncio Matter SDK API

agners:
Changes required to use the new co-routines based Matter SDK API.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -62,8 +62,6 @@ if TYPE_CHECKING:
     from collections.abc import Iterable
     from pathlib import Path
 
-    from chip.native import PyChipError
-
     from .server import MatterServer
 
 DATA_KEY_NODES = "nodes"
@@ -274,21 +272,28 @@ class MatterDeviceController:
                 attempts,
                 MAX_COMMISSION_RETRIES,
             )
-            result: (
-                PyChipError | None
-            ) = await self._chip_device_controller.commission_with_code(
-                node_id,
-                code,
-                DiscoveryType.DISCOVERY_NETWORK_ONLY
-                if network_only
-                else DiscoveryType.DISCOVERY_ALL,
-            )
-            if result and result.is_success:
-                break
-            if attempts >= MAX_COMMISSION_RETRIES:
-                raise NodeCommissionFailed(
-                    f"Commission with code failed for node {node_id}."
+            try:
+                commissioned_node_id: int = (
+                    await self._chip_device_controller.commission_with_code(
+                        node_id,
+                        code,
+                        DiscoveryType.DISCOVERY_NETWORK_ONLY
+                        if network_only
+                        else DiscoveryType.DISCOVERY_ALL,
+                    )
                 )
+                # We use SDK default behavior which always uses the commissioning Node ID in the
+                # generated NOC. So this should be the same really.
+                LOGGER.info(
+                    "Commissioned Node ID: %s vs %s", commissioned_node_id, node_id
+                )
+                assert commissioned_node_id == node_id
+                break
+            except ChipStackError as err:
+                if attempts >= MAX_COMMISSION_RETRIES:
+                    raise NodeCommissionFailed(
+                        f"Commission with code failed for node {node_id}."
+                    ) from err
             await asyncio.sleep(5)
 
         LOGGER.info("Matter commissioning of Node ID %s successful.", node_id)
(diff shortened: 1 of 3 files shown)


matter_server/server/device_controller.py
```diff
@@ -274,21 +272,28 @@ async def commission_with_code(
...
                 )
+                # We use SDK default behavior which always uses the commissioning Node ID in the
+                # generated NOC. So this should be the same really.
+                LOGGER.info(
+                    "Commissioned Node ID: %s vs %s", commissioned_node_id, node_id
+                )
+                assert commissioned_node_id == node_id
+                break
+            except ChipStackError as err:
+                if attempts >= MAX_COMMISSION_RETRIES:
```
marcelveldt, line 293: Do we still need the retries ?

matter_server/server/device_controller.py
```diff
@@ -274,21 +272,28 @@ async def commission_with_code(
...
+                        if network_only
+                        else DiscoveryType.DISCOVERY_ALL,
+                    )
                 )
+                # We use SDK default behavior which always uses the commissioning Node ID in the
+                # generated NOC. So this should be the same really.
+                LOGGER.info(
+                    "Commissioned Node ID: %s vs %s", commissioned_node_id, node_id
+                )
+                assert commissioned_node_id == node_id
```
marcelveldt, line 290: maybe raise a runtime error here or another somewhat more explicit exception instead of an assert ?
I use assert only when its 100% safe to assume the value should exist (often just to satitsfy the type checker)

matter_server/server/sdk.py
```diff
@@ -128,10 +117,9 @@ async def commission_with_code(
         node_id: int,
         setup_payload: str,
         discovery_type: DiscoveryType,
-    ) -> PyChipError:
+    ) -> int:
```
marcelveldt, line 120: So this is fixed now ? Because previously this called returned that PyChipError (which was strange enough on its own)

agners: Yeah the API was rather broken: The initial call could raise a `ChipStackError`, but the overall commissioning result was returning a `PyChipError`. This all changed with this PR https://github.com/project-chip/connectedhomeip/pull/33905.

matter_server/server/device_controller.py
agners: Yeah good question. I did many many commissionings here, in most cases the retry didn't help as the issue is something else (like incorrect code/device no longer in commissioning mode/device already commissioned with the current fabric).

However, I had one case where it indeed helped:

```
2024-06-14 12:27:18.577 (MainThread) VERBOSE [matter_server.server.client_handler] [140550928449168] Handling command commission_with_code
2024-06-14 12:27:18.578 (MainThread) INFO [matter_server.server.device_controller] Starting Matter commissioning with code using Node ID 124 (attempt 1/3).
...
2024-06-14 12:27:53.194 (Dummy-2) CHIP_PROGRESS [chip.native.DMG] Time out! failed to receive invoke command response from Exchange: 37724i
2024-06-14 12:27:53.194 (Dummy-2) CHIP_PROGRESS [chip.native.CTL] Device failed to receive the operational certificate Response: src/app/CommandSender.cpp:328: CHIP Error 0x00000032: Timeout
2024-06-14 12:27:53.194 (Dummy-2) CHIP_ERROR [chip.native.CTL] Error on commissioning step 'SendNOC': 'src/app/CommandSender.cpp:328: CHIP Error 0x00000032: Timeout'
2024-06-14 12:27:53.194 (Dummy-2) CHIP_PROGRESS [chip.native.CTL] Going from commissioning step 'SendNOC' with lastErr = 'src/app/CommandSender.cpp:328: CHIP Error 0x00000032: Timeout' -> 'Cleanup'
2024-06-14 12:27:53.195 (Dummy-2) CHIP_PROGRESS [chip.native.CTL] Performing next commissioning step 'Cleanup' with completion status = 'src/app/CommandSender.cpp:328: CHIP Error 0x00000032: Timeout'
...
2024-06-14 12:27:58.272 (MainThread) INFO [matter_server.server.device_controller] Starting Matter commissioning with code using Node ID 124 (attempt 2/3).
...
2024-06-14 12:28:10.264 (MainThread) INFO [matter_server.server.device_controller] Matter commissioning of Node ID 124 successful.
```

This was a Meross Smart Plug :man_shrugging: 

But in all other cases, it just took longer for the failure to eventually show up, which kinda hurts UX :cry: 

So I think we rather do a single one, and make sure to propage the error to the UI nicely. Maybe we can have a retry button there? :thinking:

agners: I've removed the commissioning retries. I think this is indeed more often not helpful. Let's see how this works in practice.

marcelveldt (approved):
Nice work @agners !

PR #768, opened 2024-06-21
Fix writing of Nullable attributes

marcelveldt:
Some Matter Attributes require sending a Null value to reset it to its device/factory default.
This was not working correctly, especially in the case if the annotation was a union (e/.g. an optional int).

This issue is now fixed so the write_attribute can correctly understand that it actually needs to write the sdk's Null if python sends a None.

Also added a quick test case

### DIFF
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -107,7 +107,7 @@ def parse_value(
     value: Any,
     value_type: Any,
     default: Any = MISSING,
-    allow_none: bool = True,
+    allow_none: bool = False,
     allow_sdk_types: bool = False,
 ) -> Any:
     """
@@ -121,6 +121,16 @@ def parse_value(
         # this shouldn't happen, but just in case
         value_type = get_type_hints(value_type, globals(), locals())
 
+    # handle value is None/missing but a default value is set
+    if value is None and not isinstance(default, type(MISSING)):
+        return default
+    # handle value is None and sdk type is Nullable
+    if value is None and value_type is Nullable:
+        return Nullable() if allow_sdk_types else None
+    # handle value is None (but that is allowed according to the annotations)
+    if value is None and value_type is NoneType:
+        return None
+
     if isinstance(value, dict):
         if descriptor := getattr(value_type, "descriptor", None):
             # handle matter TLV dicts where the keys are just tag identifiers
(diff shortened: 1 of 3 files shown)


tests/common/test_parser.py
```diff
@@ -110,3 +111,22 @@ def test_dataclass_from_dict():
     # test NOCStruct.noc edge case
     res = parse_value("NOCStruct.noc", 5, bytes)
     assert res == b""
+
+
+def test_parse_value():
+    """Test special cases in the parse_value helper."""
+    # ruff: noqa: PT011
```
agners, line 118: I think this is not necessary anymore

matter_server/common/helpers/util.py
```diff
@@ -107,7 +107,7 @@ def parse_value(
     value: Any,
     value_type: Any,
     default: Any = MISSING,
-    allow_none: bool = True,
+    allow_none: bool = False,
```
agners, line 110: Seems that with this PR `allow_none` is not really used anymore? Can we drop it completely?

marcelveldt: II couldnt spot where it was used but was scared to remove it

marcelveldt: it is still in use (within dataclass_from_dict)

tests/common/test_parser.py
marcelveldt: cleaned up

matter_server/common/helpers/util.py
agners: Hm, I see we pass it from `dataclass_from_dict()` to `parse_value()`. But since `dataclass_from_dict()` is in our code too, we can just drop it there no? It doesn't serve any purpose and might only lead to confusion. :man_shrugging:

marcelveldt: No, dataclass_from_dict is used in HA we cant just drop that - it does serve a purpose, parse it strict or not strict.

In strict mode a None value is seen as error if that is not in the annotations.

agners: Sorry, I didn't meant to suggest to drop `dataclass_from_dict`. I am suggesting to drop `allow_none` from parse_value()`. It needs a `dataclass_from_dict` update, but it should be a simple oneline change:


```patch
--- a/matter_server/common/helpers/util.py
+++ b/matter_server/common/helpers/util.py
@@ -272,7 +272,6 @@ def dataclass_from_dict(
                 dict_obj.get(field.name),
                 type_hints[field.name],
                 field.default,
-                allow_none=not strict,
                 allow_sdk_types=allow_sdk_types,
             )
             for field in dc_fields
```

marcelveldt: No, because that is exactly where we use it!
if the dataclass_from_dict is strict we want it to raise on None values

agners (approved):

matter_server/common/helpers/util.py
marcelveldt: Almost weekend ;-)

PR #770, opened 2024-06-24
Don't use threadsafe variant in signal_event

agners:
All calls to signal_event are made in the asyncio event loop thread. Therefore, we don't need to use the threadsafe variant of call_soon() anymore.

This was anyways a bit brittle as asyncio.create_task() is not threadsafe, so calling signal_event from a different thread was not really safe.

### DIFF
--- a/matter_server/server/server.py
+++ b/matter_server/server/server.py
@@ -252,7 +252,7 @@ class MatterServer:
             if asyncio.iscoroutinefunction(callback):
                 asyncio.create_task(callback(evt, data))
             else:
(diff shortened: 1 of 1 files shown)


agners:
Afaik, in practice this won't really make a difference as it is ok to call `call_soon_threadsafe()` from the event loop thread. But we'll get a runtime exception in the future in case we (accidentially) call `signal_event()` from another thread again.

balloob (approved):

balloob (approved):

balloob (approved):

PR #772, opened 2024-06-25
Raise InvalidServerVersion when Server is too old too

agners:
Currently InvalidServerVersion was only raised when the client was too old (Server wasn't able to handle such an old client anymore), e.g. after a backwards incompatible change Server side.

However, if the client was newer than the Server schema version, no error was raised. This essentially never triggered the add-on update logic in Core.

This change makes sure that we raise InvalidServerVersion when the Server is too old for the client too. We also add properties to the Exception so a client can figure out which situation we are dealing with (too old/too new) and act accordingly.

### DIFF
--- a/matter_server/client/connection.py
+++ b/matter_server/client/connection.py
@@ -80,13 +80,24 @@ class MatterClientConnection:
         self.server_info = info
 
         # basic check for server schema version compatibility
-        if info.min_supported_schema_version > SCHEMA_VERSION > info.schema_version:
-            # our schema version is too low and can't be handled by the server anymore.
+        if info.min_supported_schema_version > SCHEMA_VERSION:
+            # The client schema version is too low and can't be handled by the server anymore
             await self._ws_client.close()
             raise InvalidServerVersion(
                 f"Matter schema version is incompatible: {SCHEMA_VERSION}, "
                 f"the server requires at least {info.min_supported_schema_version} "
                 " - update the Matter client to a more recent version or downgrade the server.",
+                SCHEMA_VERSION,
+                info.schema_version,
+                info.min_supported_schema_version,
+            )
+        if info.schema_version < SCHEMA_VERSION:
+            # The client schema is too new and the server can't handle it yet
+            await self._ws_client.close()
+            raise InvalidServerVersion(
+                f"Matter schema version is incompatible: {SCHEMA_VERSION}, "
+                f"the server requires supports at most {info.schema_version} "
+                " - update the Matter server to a more recent version or downgrade the client.",
                 SCHEMA_VERSION,
                 info.schema_version,
                 info.min_supported_schema_version,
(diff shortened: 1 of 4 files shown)


agners:
This is an alternative to the suggested change in Core https://github.com/home-assistant/core/pull/118858.

Compared to the change in Core, it makes the connection to really fail. So this will force people to upgrade the add-on on schema bump.

matter_server/client/connection.py
```diff
@@ -80,13 +80,16 @@ async def connect(self) -> None:
         self.server_info = info
 
         # basic check for server schema version compatibility
-        if info.min_supported_schema_version > SCHEMA_VERSION:
+        if info.min_supported_schema_version > SCHEMA_VERSION > info.schema_version:
```
MartinHjelmare, line 83: How can the minimum supported schema version be higher than the latest schema version of the server? That should never happen. We need two conditions and an `or` operator.

```
Python 3.12.4 (main, Jun  8 2024, 18:29:57) [GCC 9.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.22.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: 1 > 2 > 3
Out[1]: False

In [2]: 3 > 2 > 1
Out[2]: True

In [3]: 1 > 2 > 1
Out[3]: False

In [4]: 2 > 1 > 2
Out[4]: False

In [5]: 1 > 3 > 2
Out[5]: False

In [6]: 1 > 3 or 3 > 2
Out[6]: True
```

matter_server/client/connection.py
```diff
@@ -80,13 +80,16 @@ async def connect(self) -> None:
...
             # our schema version is too low and can't be handled by the server anymore.
```
MartinHjelmare, line 84: Update this comment too.

matter_server/client/connection.py
```diff
@@ -80,13 +80,16 @@ async def connect(self) -> None:
...
             await self._ws_client.close()
             raise InvalidServerVersion(
                 f"Matter schema version is incompatible: {SCHEMA_VERSION}, "
                 f"the server requires at least {info.min_supported_schema_version} "
-                " - update the Matter client to a more recent version or downgrade the server."
+                " - update the Matter client to a more recent version or downgrade the server.",
```
MartinHjelmare, line 89: and this message

MartinHjelmare (changes requested):

matter_server/client/connection.py
agners: Hm, whoops yeah butchered that one :see_no_evil: 

I wonder if it would be better to just have two exceptions `ServerVersionTooNew`, `ServerVersionTooOld` instead of `InvalidServerVersion`? :thinking:

agners: I went down this road: There are two Exception now deriving from `InvalidServerVersion`.

MartinHjelmare (approved):
Looks good!

PR #786, opened 2024-07-02
Add flag for test-net DCL

agners:
Currently we fetch PAA certificates from the test-net DCL server by default. Often these are the same certificates as the production DCL, so this doesn't really have that much impact. However, those certs are really only required for development as certified devices will be present on the production DCL.

In the future we'll fetch more information from the DCL such as device update information etc. For these cases we really don't want production users to use test-net DCL by default.

This moves test-net DCL use behind a flag. By default this should be off.

### DIFF
--- a/matter_server/server/__main__.py
+++ b/matter_server/server/__main__.py
@@ -98,6 +98,12 @@ parser.add_argument(
     default=None,
     help="Directory where PAA root certificates are stored.",
 )
+parser.add_argument(
+    "--use-test-net-dcl",
+    type=bool,
+    default=False,
+    help="Use PAA root certificates and other device information from test-net DCL.",
+)
 
 args = parser.parse_args()
 
@@ -181,6 +187,7 @@ def main() -> None:
         args.listen_address,
         args.primary_interface,
         args.paa_root_cert_dir,
+        args.use_test_net_dcl,
     )
 
     async def handle_stop(loop: asyncio.AbstractEventLoop) -> None:
--- a/matter_server/server/__main__.py
+++ b/matter_server/server/__main__.py
@@ -99,10 +99,10 @@ parser.add_argument(
     help="Directory where PAA root certificates are stored.",
 )
 parser.add_argument(
-    "--use-test-net-dcl",
+    "--enable-test-net-dcl",
     type=bool,
     default=False,
-    help="Use PAA root certificates and other device information from test-net DCL.",
+    help="Enable PAA root certificates and other device information from test-net DCL.",
 )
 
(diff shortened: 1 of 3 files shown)


marcelveldt:
I'd like this to be splitted; so an explicit flag for using test DCL for certificates and one for device updates.
For device updates, this is mainly a request from matter device developers and folks involved with matter/device development to be able to test the OTA from test DCL for a particular device(range). 

So I'm proposing we add a way to explicitly allow test DCL on a vendor or product id base.
That reduces the risk while still allowing developers (including us) to testdrive their updates.

agners:
I don't like splitting as there will be more things coming in the future, like checking certification status etc.

So that would add yet another flag. IMHO, it's not worth to have a "flag forest" just for test-net DCL things :man_shrugging: 

> So I'm proposing we add a way to explicitly allow test DCL on a vendor or product id base.

This would add even more fields, just for testing. Even if we poll test-net DCL for device I am currently not interested, does it matter all that much? Internet is fast these days :man_shrugging:  

And if you are not developer, don't enable it.

> That reduces the risk while still allowing developers (including us) to testdrive their updates.

Sure this will potentially show updates for devices we are not interested. But we don't auto-update, so just don't click that button no? :sweat_smile: 

Btw, on my desk:
- Breakages from main net 1
- Breakages from test net 0 (even though I did much more updates from here :see_no_evil: )

marcelveldt:
OK, lets settle on a 2 phase step then:

One toggle enables test DCL but keep the explicit list for updates from test DCL.
I am just really worried that folks enable test DCL and then accidentally update to a test firmware on test DCL so an additional filter is really needed here.

matter_server/server/__main__.py
```diff
@@ -98,6 +98,12 @@
     default=None,
     help="Directory where PAA root certificates are stored.",
 )
+parser.add_argument(
+    "--use-test-net-dcl",
```
marcelveldt, line 102: maybe the flag should be "allow-test-net-dcl" or "enable-test-net-dcl" because strictly taken its using both main and test with the toggle enabled.

agners:
> OK, lets settle on a 2 phase step then:

Ok, works for me. 

> One toggle enables test DCL but keep the explicit list for updates from test DCL.

I'll implement this part with the update PR then.

matter_server/server/__main__.py
agners: Ok, I go for `enable-test-net-dcl`.

marcelveldt (approved):
Nice!

PR #798, opened 2024-07-11
A small optimization to the node setup

marcelveldt:
If you have issues in your Thread network and/or multiple misbehaving devices we get trapped into a deadlock situation if we have 5 nodes stuck in setup. In that case no devices will be setup at all.

This changes the logic a tiny bit so we still use a semaphore to somewhat throttle the burst of traffic that the setup generates on the thread network but back out of it in case a node hangs in there.

- Use set-up semaphore only for Thread devices
- Release semaphore when we detect a node hangs in setup

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -132,7 +132,7 @@ class MatterDeviceController:
         self._aiozc: AsyncZeroconf | None = None
         self._fallback_node_scanner_timer: asyncio.TimerHandle | None = None
         self._fallback_node_scanner_task: asyncio.Task | None = None
-        self._node_setup_throttle = asyncio.Semaphore(5)
+        self._thread_node_setup_throttle = asyncio.Semaphore(5)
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -1164,52 +1194,30 @@ async def log_node_long_setup(time_start: float) -> None:
...
+                self._schedule_custom_attributes_poller()
 
-            finally:
-                log_timers[node_id].cancel()
-                self._nodes_in_setup.discard(node_id)
+        finally:
+            log_timers[node_id].cancel()
+            self._nodes_in_setup.discard(node_id)
+            if is_thread_node:
+                self._thread_node_setup_throttle.release()
```
agners, line 1220: We potentially double release a semaphore here when the node was stuck previously but then gets unstuck no?

marcelveldt: Yeah, correct but that doesn't seem to harm as its just a counter

agners (approved):

PR #800, opened 2024-07-11
Add node id to log records

agners:
Create a Matter specific logging formatter which formats log records with node information differently than other log records. This makes following a particular nodes log entries easier.

Also add a new argument which allows to filter logs by Nodes. This allows to show only the logs of a particular set of node IDs.


matter_server/server/__main__.py
```diff
@@ -103,6 +104,12 @@
     action="store_true",
     help="Enable PAA root certificates and other device information from test-net DCL.",
 )
+parser.add_argument(
+    "--log-node-ids",
+    type=int,
+    nargs="+",
+    help="List of node IDs to show logs from (applies only to server logs).",
+)
```
marcelveldt, line 112: Can we somehow turn this into a runtime config ?

matter_server/server/sdk.py
```diff
@@ -365,7 +374,7 @@ async def find_or_establish_case_session(
         if self._chip_controller is None:
             raise RuntimeError("Device Controller not initialized.")
 
-        node_logger = LOGGER.getChild(f"node_{node_id}")
+        node_logger = self.get_node_logger(LOGGER, node_id)
```
marcelveldt, line 377: Now it would be really awesome if we could somehow also redirect the chip sdk native logs that include the node ID (now in hex format) because that is the most annoying to keep track of in the jungle of log output. Guess that is one bridge too far ?

agners: Yeah I did think about that. A hacky method to do that would be to parse the log string. However, the log strings are not uniform (look a bit different at times), so this is potentially unreliable. Also, performance wise it isn't ideal as the logs would bubble up from the native side to the Python side (acquiring the GIL), to then only getting filtered out there.

Ideally the SDK would support node id log filtering as well, which we then could set similarly.

I think the second is a bit too much work, so maybe it is still worth doing the first one for now. But this would be a follow up PR.

matter_server/server/__main__.py
marcelveldt: I was thinking of exposing it as ws command and make a nice button in the debug UI

agners: The downside of that is logging during startup then is unfiltered. I kinda like the add-on config approach since it sets up the filter at the very start.

The feature also might be used by relatively non-technical people. I'd prefer if we don't have to lead them to the debug UI just to get a startup log for a particular problematic node. 

How about adding the runtime log filter in a follow up?

marcelveldt (approved):

PR #805, opened 2024-07-16
Matter update fixes

agners:
Make sure parent directories get created to when creating the node specific update directory.

Fix save/restore of the UpdateSource.

### DIFF
--- a/matter_server/server/ota/provider.py
+++ b/matter_server/server/ota/provider.py
@@ -82,7 +82,10 @@ class ExternalOtaProvider:
         loop = asyncio.get_event_loop()
 
         await loop.run_in_executor(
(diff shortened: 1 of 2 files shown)


matter_server/server/ota/provider.py
```diff
@@ -82,7 +82,10 @@ async def initialize(self) -> None:
         loop = asyncio.get_event_loop()
 
         await loop.run_in_executor(
-            None, functools.partial(self._ota_provider_dir.mkdir, exist_ok=True)
+            None,
+            functools.partial(
+                self._ota_provider_dir.mkdir, exist_ok=True, parents=True
+            ),
         )
```
marcelveldt, line 90: you could also replace this with asyncio.to_thread

marcelveldt (approved):

PR #811, opened 2024-07-18
Skip updates from DCL which are considered invalid

agners:
Skip updates which have the softwareVersionValid flag set to false. This avoids flashing potentially invalid software versions to Matter devices.


marcelveldt (approved):
LGTM

PR #827, opened 2024-07-22
Log when the Matter server finished startup

agners:
Add a log entry with info level when the Matter server finishes startup. This is useful to know when the server is ready and starts event based Node setup processing (from mDNS events).

Helps for situation like https://github.com/home-assistant/core/issues/121932

### DIFF
--- a/matter_server/server/server.py
+++ b/matter_server/server/server.py
@@ -207,7 +207,7 @@ class MatterServer:
             self._runner, host=self.listen_addresses, port=self.port
         )
         await self._http.start()
-        self.logger.debug("Webserver initialized.")
+        self.logger.info("Matter Server successfully initialized.")
 
     async def stop(self) -> None:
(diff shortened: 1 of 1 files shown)


marcelveldt:
Do we also want to add a log line with the number of nodes we have in storage ?

Something like "Starting mdns disocvery for X known nodes"

agners:
> Do we also want to add a log line with the number of nodes we have in storage ?
> 
> Something like "Starting mdns disocvery for X known nodes"

We have that, shortly before:

```
2024-07-14 13:04:02.983 (MainThread) INFO [matter_server.server.device_controller] Loaded 3 nodes from stored configuration
```

It is less noticable because vendor info/DCL gets loaded after that.

Probably make sense to order it to do that after loading DCL/vendor info.

marcelveldt (approved):

PR #829, opened 2024-07-22
Don't regard update information without a valid URL

agners:
Skip updates which don't have an URL. We can't update such a node, so simply pretend as if they don't exist in the version info.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1024,7 +1024,7 @@ class MatterDeviceController:
             node_logger.info("No new update found.")
             return None, None
 
-        if "otaUrl" not in update:
+        if "otaUrl" not in update or update["otaUrl"].strip().length == 0:
             raise UpdateCheckError("Update found, but no OTA URL provided.")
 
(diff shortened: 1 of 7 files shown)


matter_server/server/device_controller.py
```diff
@@ -1024,7 +1024,7 @@ async def _check_node_update(
             node_logger.info("No new update found.")
             return None, None
 
-        if "otaUrl" not in update:
+        if "otaUrl" not in update or update["otaUrl"].trim().length == 0:
```
marcelveldt, line 1027:
```suggestion
        if "otaUrl" not in update or update["otaUrl"].strip().length == 0:
```

agners: Actually, this check should not be there :see_no_evil:  I moved it to the `_check_update_version()` function.

Well, I guess we could also drop it here, in case someone adds a local udpate with an empty URL :thinking:

agners: Fixed!

marcelveldt (approved):

PR #855, opened 2024-08-12
Eve cluster part2: Add EveWeather attributes

lboue:
Add Class attribute for Eve Cluster (EveWeather device):
- Altitude
- Pressure

HA core PR: [Add EveWeatherPressure attribute #123565](https://github.com/home-assistant/core/pull/123565)

### DIFF
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -291,6 +291,7 @@ class EveCluster(Cluster, CustomClusterMixin):
 
             value: int = 0
 
+
 @dataclass
 class NeoCluster(Cluster, CustomClusterMixin):
     """Custom (vendor-specific) cluster for Neo - Vendor ID 4991 (0x137F)."""
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -99,6 +99,8 @@ class EveCluster(Cluster, CustomClusterMixin):
     wattAccumulatedControlPoint: float32 | None = None
     voltage: float32 | None = None
     current: float32 | None = None
+    altitude: int | None = None
+    pressure: float32 | None = None
 
     class Attributes:
         """Attributes for the Eve Cluster."""
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -266,7 +266,7 @@ class EveCluster(Cluster, CustomClusterMixin):
                 """Return attribute type."""
                 return ClusterObjectFieldDescriptor(Type=int)
 
-            value: int = 0
+            value: float32 = 0
 
         @dataclass
         class Pressure(ClusterAttributeDescriptor, CustomClusterAttributeMixin):
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -249,8 +249,6 @@ class EveCluster(Cluster, CustomClusterMixin):
         class Altitude(ClusterAttributeDescriptor, CustomClusterAttributeMixin):
             """Altitude Attribute within the Eve Cluster."""
 
-            should_poll = False
-
             @ChipUtility.classproperty
             def cluster_id(cls) -> int:
                 """Return cluster id."""
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -85,7 +85,7 @@ class EveCluster(Cluster, CustomClusterMixin):
                     Label="current", Tag=0x130A0009, Type=float32
                 ),
                 ClusterObjectFieldDescriptor(
-                    Label="altitude", Tag=0x130A0013, Type=int
+                    Label="altitude", Tag=0x130A0013, Type=float32
                 ),
                 ClusterObjectFieldDescriptor(
                     Label="pressure", Tag=0x130A0014, Type=float32
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -99,7 +99,6 @@ class EveCluster(Cluster, CustomClusterMixin):
     wattAccumulatedControlPoint: float32 | None = None
     voltage: float32 | None = None
     current: float32 | None = None
-    altitude: int | None = None
     pressure: float32 | None = None
 
     class Attributes:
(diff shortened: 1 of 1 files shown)


agners (approved):

matter_server/common/custom_clusters.py
```diff
@@ -243,6 +245,52 @@ def attribute_type(cls) -> ClusterObjectFieldDescriptor:
 
             value: float32 = 0
 
+        @dataclass
+        class Altitude(ClusterAttributeDescriptor, CustomClusterAttributeMixin):
+            """Altitude Attribute within the Eve Cluster."""
+
+            should_poll = True
```
agners, line 252: Are you sure this need to be polled as well? :thinking:

lboue: No, that's not necessary.

agners: I didn't meant to remove altitude, just use the subscription mechanism (if subscriptions work on these attributes, that you'd have to test on an actual device).

lboue: How can I manually test the subscription mechanism with my device?

agners: You can just set the log level to verbose, then you'll see what is happening. The `Attribute updated` string comes from the subscription callback: 

https://github.com/home-assistant-libs/python-matter-server/blob/c09db74df9f6ce43a37cff88a148069a0c15fd50/matter_server/server/device_controller.py#L1056-L1067

I actually bought such a device too, and I can see that the altitude gets updated via the subscription:
```
2024-08-16 14:39:29.781 (MainThread) VERBOSE [matter_server.server.device_controller] <Node:264> Attribute updated: 1/319486977/319422484 - old value: 959.7000122070312 - new value: 959.0999755859375
```

So in other words, we can set polled to false here, and it will still update :tada:

matter_server/common/custom_clusters.py
```diff
@@ -243,6 +245,52 @@ def attribute_type(cls) -> ClusterObjectFieldDescriptor:
...
+
+            @ChipUtility.classproperty
+            def attribute_id(cls) -> int:
+                """Return attribute id."""
+                return 0x130A0013
+
+            @ChipUtility.classproperty
+            def attribute_type(cls) -> ClusterObjectFieldDescriptor:
+                """Return attribute type."""
+                return ClusterObjectFieldDescriptor(Type=int)
```
agners, line 267: Actually, it seems that this attribute is a float32 (from my logs):

```
TypeError: Value 0.0 of type <class 'float'> is invalid for altitude, expected value of type <class 'int'>
```

However, I see that the value currently stays at `0`. I don't think it is useful currently, so I think it is better to  just drop it until it actually provides useful information. Sorry for the back and fourth.

matter_server/common/custom_clusters.py
```diff
@@ -243,6 +245,52 @@ def attribute_type(cls) -> ClusterObjectFieldDescriptor:
...
+
+            value: int = 0
+
+        @dataclass
+        class Pressure(ClusterAttributeDescriptor, CustomClusterAttributeMixin):
+            """Pressure Attribute within the Eve Cluster."""
+
+            should_poll = True
```
agners, line 275: As discussed in the other review, this does not need to be polled. The default is `False`, you can just drop it:

```suggestion
```

agners (changes requested):

matter_server/common/custom_clusters.py
lboue: In fact, it was set to zero because I hadn't configured it in the Eve application in the first place. Since then, the value is different from `0`

agners: Uh I see, this is more a configuration value :thinking: Yeah that make sense, it is probably used to calibrate/improve read out of the pressure sensor.

Ideally we would add it as a configuration then to HA.

lboue: image

matter_server/common/custom_clusters.py
```diff
@@ -99,6 +99,8 @@ def descriptor(cls) -> ClusterObjectDescriptor:
     wattAccumulatedControlPoint: float32 | None = None
     voltage: float32 | None = None
     current: float32 | None = None
+    altitude: int | None = None
```
agners, line 102: Altitude is in fact float32 too, so:

```suggestion
    altitude: float32 | None = None
```

matter_server/common/custom_clusters.py
```diff
@@ -243,6 +245,50 @@ def attribute_type(cls) -> ClusterObjectFieldDescriptor:
...
+
+            @ChipUtility.classproperty
+            def attribute_id(cls) -> int:
+                """Return attribute id."""
+                return 0x130A0013
+
+            @ChipUtility.classproperty
+            def attribute_type(cls) -> ClusterObjectFieldDescriptor:
+                """Return attribute type."""
+                return ClusterObjectFieldDescriptor(Type=int)
```
agners, line 267:
```suggestion
                return ClusterObjectFieldDescriptor(Type=float32)
```

matter_server/common/custom_clusters.py
```diff
@@ -243,6 +245,50 @@ def attribute_type(cls) -> ClusterObjectFieldDescriptor:
...
+
+            value: int = 0
```
agners, line 269:
```suggestion
            value: float32 = 0
```

agners (changes requested):
And finally, you also need to adjust it here: https://github.com/home-assistant-libs/python-matter-server/blob/1c3814722070295bfb349bb50b4fc81a8b7b3e1a/matter_server/common/custom_clusters.py#L88

lboue:
> And finally, you also need to adjust it here:
> 
> https://github.com/home-assistant-libs/python-matter-server/blob/1c3814722070295bfb349bb50b4fc81a8b7b3e1a/matter_server/common/custom_clusters.py#L88

Fixed.

matter_server/common/custom_clusters.py
```diff
@@ -243,6 +245,50 @@ def attribute_type(cls) -> ClusterObjectFieldDescriptor:
 
             value: float32 = 0
 
+        @dataclass
+        class Altitude(ClusterAttributeDescriptor, CustomClusterAttributeMixin):
+            """Altitude Attribute within the Eve Cluster."""
+
+            should_poll = False
+
```
agners, line 253: We can drop this too.

```suggestion
```

agners (approved):
Other than this, LGTM! Thanks!

lboue:
Working fine with [6.5.0 BETA 1](https://github.com/home-assistant-libs/python-matter-server/releases/tag/6.5.0b1) version:

image

PR #873, opened 2024-09-03
Retry subscription setup if necessary

agners:
With the update of the Matter SDK wheels to 2024.7.1 (#809) the SDK no longer auto-resubscribes when the read of the subscription fails (see https://github.com/project-chip/connectedhomeip/pull/34370). The Server had a comment that it relies on mDNS rediscovery to re-establish the subscription. But this is not always the case. This change adds a retry mechanism to the subscription setup for a fixed period of time after seeing it via mDNS.

Additionally, it avoids recreating the subscription if the liveness timeout did not occur yet. This puts more trust into the SDK handling the subscriptions.

Fixes: https://github.com/home-assistant/core/issues/124503

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -144,7 +144,7 @@ class MatterDeviceController:
         self._fabric_id_hex: str | None = None
         self._wifi_credentials_set: bool = False
         self._thread_credentials_set: bool = False
-        self._nodes_in_setup: set[int] = set()
+        self._setup_node_with_retry_tasks = dict[int, asyncio.Task]()
         self._nodes_in_ota: set[int] = set()
         self._node_last_seen_on_mdns: dict[int, float] = {}
         self._nodes: dict[int, MatterNodeData] = {}
@@ -230,6 +230,8 @@ class MatterDeviceController:
             scan_task.cancel()
         if self._aiozc:
             await self._aiozc.async_close()
+        for task in self._setup_node_with_retry_tasks.values():
+            task.cancel()
 
         # shutdown the sdk device controller
         await self._chip_device_controller.shutdown()
@@ -340,7 +342,8 @@ class MatterDeviceController:
                 break
 
         # make sure we start a subscription for this newly added node
-        await self._setup_node_with_retry(node_id)
+        if task := self._setup_node_create_task(node_id):
+            await task
         LOGGER.info("Commissioning of Node ID %s completed.", node_id)
         # return full node object once we're complete
         return self.get_node(node_id)
@@ -418,7 +421,8 @@ class MatterDeviceController:
             else:
                 break
         # make sure we start a subscription for this newly added node
-        await self._setup_node_with_retry(node_id)
+        if task := self._setup_node_create_task(node_id):
+            await task
         LOGGER.info("Commissioning of Node ID %s completed.", node_id)
         # return full node object once we're complete
         return self.get_node(node_id)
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -1383,21 +1371,51 @@ async def log_node_long_setup(time_start: float) -> None:
...
+                    await self._setup_node(node_logger, node_id)
+                    break
+                except (NodeNotResolving, NodeInterviewFailed, ChipStackError):
+                    if (
+                        time.time() - self._node_last_seen_on_mdns.get(node_id, 0)
+                        > NODE_MDNS_SUBSCRIPTION_RETRY_TIMEOUT
+                    ):
+                        # NOTE: assume the node will be picked up by mdns discovery later
+                        # automatically when it becomes available again.
+                        logging.info(
```
marcelveldt, line 1408:
```suggestion
                        logging.warning(
```

matter_server/server/device_controller.py
```diff
@@ -1511,7 +1524,7 @@ def _on_mdns_operational_node_state(
             return
 
         # setup the node - this will (re) setup the subscriptions etc.
-        asyncio.create_task(self._setup_node(node_id))
+        asyncio.create_task(self._setup_node_with_retry(node_id))
```
marcelveldt, line 1527: Let's store a reference to this task in the device controller. Make self._nodes_in_setup a dict that stores the node id as key and the task as value - that way we can also safely shutdown all these setup tasks at shutdown

matter_server/server/device_controller.py
```diff
@@ -1494,9 +1510,6 @@ def _on_mdns_operational_node_state(
 
         if not self._chip_device_controller.node_has_subscription(node_id):
             node_logger.info("Discovered on mDNS")
-        elif (now - last_seen) > NODE_MDNS_BACKOFF:
-            # node came back online after being offline for a while or restarted
-            node_logger.info("Re-discovered on mDNS")
```
marcelveldt, line 1499: We need to properly test this - some magic things can happen if a device reboots within the backoff period or outside our mdns window and that results in a burst of removed and added events

matter_server/server/device_controller.py
```diff
@@ -1383,21 +1371,51 @@ async def log_node_long_setup(time_start: float) -> None:
...
+            raise NodeNotExists(f"Node {node_id} does not exist.")
+        if node_id in self._nodes_in_setup:
+            # prevent duplicate setup actions
+            return
+        self._nodes_in_setup.add(node_id)
+
+        node_logger = self.get_node_logger(LOGGER, node_id)
+
+        try:
+            while True:
```
marcelveldt, line 1397: instead of this nested try and while loop you could also schedule a timer to retry the task after 60 seconds and abort that timerhandle.

matter_server/server/device_controller.py
agners: The removed events don't make it here, so we are good on these.

As for the Updated and Added events, we process fewer of those now: Only if there is no subscription (which is the same as before) we go into node_setup. So this certainly leads to less node_setup calls.

But it will more often call into the next branch, and call `trigger_resubscribe_if_scheduled`. But this doesn't do much other than setting a flag and forcing a timer timeout on SDK side.

matter_server/server/device_controller.py
agners: We do use call setup node from other places too, without create_task. So this needs a bit more refactoring :thinking:

matter_server/server/device_controller.py
agners: Yeah, in the end technically it doesn't matter much. It is work on the asyncio loop which will get executed delayed. I feel the while loop makes it much more readable and easier to reason when retrying actually stops.

matter_server/server/device_controller.py
marcelveldt: yes but now you are suggesting a long(er) running task

matter_server/server/device_controller.py
marcelveldt: It doesn't matter - its just a case of preference. Let's leave it this way - if we find it unreadable later we can always make it more pretty. One thing we need to do though is store the reference to the task

matter_server/server/device_controller.py
agners: I am creating a task always now, so we can cancel any running setup at the end.

matter_server/server/device_controller.py
```diff
@@ -230,6 +230,8 @@ async def stop(self) -> None:
             scan_task.cancel()
         if self._aiozc:
             await self._aiozc.async_close()
+        for task in self._setup_node_with_retry_tasks.values():
+            task.cancel()
```
marcelveldt, line 234:
```suggestion
        for task in self._setup_node_with_retry_tasks.values():
            if not task.done():
                task.cancel()
```

matter_server/server/device_controller.py
```diff
@@ -1383,21 +1375,54 @@ async def log_node_long_setup(time_start: float) -> None:
...
+            node_logger.info("Retrying node setup in 60 seconds...")
+            await asyncio.sleep(60)
+
+    def _setup_node_create_task(self, node_id: int) -> asyncio.Task | None:
+        """Create a task for setting up a node with retry."""
+        if node_id in self._setup_node_with_retry_tasks:
+            node_logger = self.get_node_logger(LOGGER, node_id)
+            node_logger.debug("Setup task exists already for this Node")
+            return None
+        task = asyncio.create_task(self._setup_node_with_retry(node_id))
```
marcelveldt, line 1422: bit of an edge case maybe but we should also ensure this task gets cancelled in the remove node function

matter_server/server/device_controller.py
agners: `task.cancel()` is safe to be called for cancelled tasks. That said, maybe we should remove the tasks once they are done? :thinking:  On the other hand, why bother...

marcelveldt: I was just wondering the same, should we cleanup the list and pop an item out if its done.

marcelveldt (approved):
Looks good!

Drealine:
Still got the same issue for me : 

`[139815571515872] Unexpected exception
[139815779709520] Unexpected exception
[139815459890672] Unexpected exception
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 245, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2761, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2804, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 996, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1068, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/lock/__init__.py", line 226, in async_handle_lock_service
    await self.async_lock(**self.add_default_code(kwargs))
  File "/usr/src/homeassistant/homeassistant/components/matter/lock.py", line 90, in async_lock
    await self.send_device_command(
  File "/usr/src/homeassistant/homeassistant/components/matter/lock.py", line 70, in send_device_command
    await self.matter_client.send_device_command(
  File "/usr/local/lib/python3.12/site-packages/matter_server/client/client.py", line 454, in send_device_command
    return await self.send_command(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/matter_server/client/client.py", line 583, in send_command
    return await future
           ^^^^^^^^^^^^
matter_server.common.errors.UnknownError: InteractionModelError: UnsupportedAccess (0x7e)`

PR #876, opened 2024-09-04
Shutdown subscription after exactly 30 minutes

agners:
Instead of relying on the progressively growing and random resubscribe interval simply check the time since the first re-subscription attempt.

The resubscribe interval is chosen at random, and even in the terminal state, a resubscription potentially happens below 30 minutes (in the final progression the resubscription time is chosen at random between 27 and 92 minutes). This means that in theory we shutdown the subscription much later than actually intended.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -1206,7 +1206,7 @@ class MatterDeviceController:
                 self._node_unavailable(node_id)
             # Shutdown the subscription if we tried to resubscribe for more than 30
             # minutes (typical TTL of mDNS). We assume this device got powered off.
-            # When the device gets powered on again, it typically announced itself via
+            # When the device gets powered on again, it typically announces itself via
             # mDNS again. The mDNS browsing code will setup the subscription again.
             if (
                 time.time() - self._first_resubscribe_attempt[node_id]
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -150,7 +150,7 @@ class MatterDeviceController:
         self._node_last_seen_on_mdns: dict[int, float] = {}
         self._nodes: dict[int, MatterNodeData] = {}
         self._last_known_ip_addresses: dict[int, list[str]] = {}
-        self._last_subscription_attempt: dict[int, int] = {}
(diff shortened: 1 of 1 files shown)


agners:
It seems that the callback provided time is "made up", and doesn't reflect the actual resubscription time, e.g.:

```
2024-09-04 10:20:59.046 (Dummy-2) CHIP_PROGRESS [chip.native.DMG] Will try to resubscribe to 01:0000000000000001 at retry index 5 after 29988ms due to error src/protocols/secure_channel/CASESession.cpp:560: CHIP Error 0x00000032: Timeout
2024-09-04 10:20:59.047 (MainThread) INFO [matter_server.server.device_controller] <Node:1> Previous subscription failed with Error: 50, re-subscribing in 21192 ms...
...
2024-09-04 10:21:29.035 (Dummy-2) CHIP_PROGRESS [chip.native.DMG] OnResubscribeTimerCallback: ForceCASE = 1
```

(the actual resubscribe time was after ~29.9s, so the value provided by the SDK, not the Python callback)

This change just removes the Python callback provided time, it is useless currently.

matter_server/server/device_controller.py
```diff
@@ -1188,29 +1190,37 @@ def resubscription_attempted(
...
-            # after some resubscription attempts and we shutdown the subscription
-            # if the resubscription interval exceeds 30 minutes (TTL of mdns).
-            # The node will be auto picked up by mdns if it's alive again.
+            # after some resubscription attempts.
             if resubscription_attempt >= NODE_RESUBSCRIBE_ATTEMPTS_UNAVAILABLE:
                 self._node_unavailable(node_id)
-            if nextResubscribeIntervalMsec > NODE_RESUBSCRIBE_TIMEOUT_OFFLINE:
+            # Shutdown the subscription if we tried to resubscribe for more than 30
+            # minutes (typical TTL of mDNS). We assume this device got powered off.
+            # When the device gets powered on again, it typically announced itself via
```
marcelveldt, line 1209:
```suggestion
            # When the device gets powered on again, it typically announces itself via
```

marcelveldt (approved):

PR #882, opened 2024-09-12
Cleanup: Remove fallback node scanner

agners:
The fallback node scanner introduced with #603 intents to pick up nodes which our zeroconf scanner missed. There is currently no indication that the zeroconf scanner misses nodes (anymore), so the fallback scanner is redundant. This PR removes the fallback scanner.


agners:
I've searched for logs with `discovered using fallback ping`. There are some reports, such as [this community post](https://community.home-assistant.io/t/aqara-p2-sensors-status-changes-to-unavailable/739099/4). But it seems that in this case the same node has been discovered through mDNS as well, but just refuses to setup correctly. So I don't think the fallback does add any value today, especially since we have #873 in place.

marcelveldt (approved):

PR #885, opened 2024-09-12
Select OTA Provider port randomly

agners:
Instead of using a fixed port for OTA Provider, pass 0 to let the Matter SDK/bind() syscall select a random port. This allows multiple OTA Providers to run on the same host without port conflicts.

Fixes: https://github.com/home-assistant/core/issues/125130


agners:
Pretty sure I had dynamic ports used at one point, but I think I was unsure if issues we've seen with Apple Thread border router were maybe related to port selection. Anyhow, initial investigation point more towards mDNS issue on that front.

marcelveldt (approved):

PR #886, opened 2024-09-13
Use to_thread where appropriate

agners:
Use asyncio.to_thread where appropriate, and replace existing loop.run_in_executor() calls with the shorter asyncio.to_thread() call.


MartinHjelmare (approved):
Thanks!

PR #892, opened 2024-09-19
Use interval floor of 0 for ICD devices

marcelveldt:
Yesterday we updated the mininterval floor to 1 second, to prevent a flood of updates at an insane frequency (e.g. transitions) but apparently this is not recommended for ICD devices.

So small follow-up to use 0 for ICD and 1 second for all other.

Scherm­afbeelding 2024-09-19 om 17 03 41

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -117,6 +117,9 @@ BASIC_INFORMATION_SOFTWARE_VERSION_STRING_ATTRIBUTE_PATH = (
         0, Clusters.BasicInformation.Attributes.SoftwareVersionString
     )
 )
+ICD_ATTR_LIST_ATTRIBUTE_PATH = create_attribute_path_from_attribute(
+    0, Clusters.IcdManagement.Attributes.AttributeList
+)
 
 
 # pylint: disable=too-many-lines,too-many-instance-attributes,too-many-public-methods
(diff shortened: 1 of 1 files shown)


matter_server/server/device_controller.py
```diff
@@ -1236,6 +1239,13 @@ def resubscription_succeeded(
             interval_ceiling = NODE_SUBSCRIPTION_CEILING_BATTERY_POWERED
         else:
             interval_ceiling = NODE_SUBSCRIPTION_CEILING_THREAD
+        if node.attributes.get(ICD_ATTR_LIST_ATTRIBUTE_PATH) is not None:
+            # for ICD devices, the interval floor must be 0 according to the spec,
+            # to prevent additional battery drainage. See Matter core spec, chapter 8.5.2.2.
+            # TODO: revisit this after Matter 1.4 release (as that mighht change this again).
+            interval_floor = 0
```
agners, line 1246: Can we use a variable here too? `NODE_SUBSCRIPTION_FLOOR_ICD` or so?

marcelveldt: done

agners (approved):

bjalt:
@marcelveldt Doesn't `NODE_SUBSCRIPTION_FLOOR_ICD = 1` and the use of this variable set the MinFloorValue to `1` instead of the intended value of `0`?

marcelveldt:
> @marcelveldt Doesn't `NODE_SUBSCRIPTION_FLOOR_ICD = 1` and the use of this variable set the MinFloorValue to `1` instead of the intended value of `0`?

yeah, good catch! We missed that completely, not enough coffee I guess ;-)
Thanks!

PR #893, opened 2024-09-19
Prevent Eve devices with newer firmware to be polled

marcelveldt:
The server has some logic to poll custom attributes when needed, this is for example needed for the custom power/energy attributes on Eve Energy with firmware version < 3.5.

The newer Eve Energy firmware however implements the official Matter 1.3 clusters for power and energy measurements so no need to to poll these attributes anymore.

- Changes should_poll in the custom cluster/attribute definitions into a method/callback that accepts the MatterNodeData as argument. That way the implementation can do some conditional checks based on node attribute/firmware version etc.

- Disable polling if the node is not Eve - this catches e.g. the MatterBridge which (ab)uses the Eve custom cluster to send power measurements. There is no reason to poll those values.

- Disable polling on the Eve power attributes if the node has the new ElectricalPowerMeasurement cluster.

Strictly taken this means that you need to update HA in order to use the new sensors but since the Eve firmware is not yet available this is not a real problem. I could have gone with a super dynamic client-dependent approach where the client determines what attributes should be polled but imo that is overkill for these few attributes and the fact that we're just preparing for the future here. Home Assistant 2024.9 will have the new power sensors available.

### DIFF
--- a/matter_server/common/custom_clusters.py
+++ b/matter_server/common/custom_clusters.py
@@ -62,6 +62,17 @@ class CustomClusterAttributeMixin:
         ALL_CUSTOM_ATTRIBUTES[cls.cluster_id][cls.attribute_id] = cls
 
 
+def should_poll_eve_energy(node_data: MatterNodeData) -> bool:
+    """Check if the (Eve Energy) custom attribute should be polled for state changes."""
+    if node_data.attributes.get("0/40/2") != 4874:
+        # Some implementation (such as MatterBridge) use the
+        # Eve cluster to send the power measurements. Filter that out.
+        return False
+    # if the ElectricalPowerMeasurement cluster is NOT present,
+    # we should poll the custom Eve cluster attribute(s).
+    return node_data.attributes.get("2/144/65531") is None
+
+
 @dataclass
 class EveCluster(Cluster, CustomClusterMixin):
     """Custom (vendor-specific) cluster for Eve - Vendor ID 4874 (0x130a)."""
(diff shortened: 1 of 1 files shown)


matter_server/common/custom_clusters.py
```diff
@@ -55,6 +62,17 @@ def __init_subclass__(cls: ClusterAttributeDescriptor, *args, **kwargs) -> None:
...
 
+def should_poll_eve_energy(node_data: MatterNodeData) -> bool:
+    """Check if the (Eve Energy) custom attribute should be polled for state changes."""
+    if node_data.attributes.get("0/40/2") != 4874:
+        # Some implementation (such as MatterBridge) use the
+        # Eve cluster to send the power measurements. Filter that out.
+        return False
+    # if the ElectricalPowerMeasurement cluster is NOT present,
+    # we should poll the custom Eve cluster attribute(s).
+    return node_data.attributes.get("2/144/65531") is None
```
agners, line 73: In what order do things happen after update? Do we get an attribute updates first before we check what needs polled? :thinking:

matter_server/common/custom_clusters.py
```diff
@@ -55,6 +62,17 @@ def __init_subclass__(cls: ClusterAttributeDescriptor, *args, **kwargs) -> None:
         ALL_CUSTOM_ATTRIBUTES[cls.cluster_id][cls.attribute_id] = cls
 
 
+def should_poll_eve_energy(node_data: MatterNodeData) -> bool:
+    """Check if the (Eve Energy) custom attribute should be polled for state changes."""
+    if node_data.attributes.get("0/40/2") != 4874:
```
agners, line 67: I don't like these "magic numbers" too much, if we can have something which is based on the cluster library it would be much more readable what is happening here (and also below). :thinking:

matter_server/common/custom_clusters.py
marcelveldt: I think we have to assume that this will be at random. So worst case scenario we still poll the custom value after a device gets updated. That is no big deal. At some point when the matter server is restarted (e.g. after an update) it will stop polling. Sure we could try to handle that (edge) case but to what purpose other than writing even more complicated code.

matter_server/common/custom_clusters.py
marcelveldt: Well, yes and no. At least its really consolidated to one single function here for Eve.
But I'll create some constants

marcelveldt: adjusted

MartinHjelmare (approved):

PR #899, opened 2024-09-27
Add fabric label on commissioning

agners:
Add a fabric label right after commissioning. This will make it easier for other platforms to identify what particular fabrics a device is commissioned with.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -166,7 +166,7 @@ class MatterDeviceController:
         self._custom_attribute_poller_timer: asyncio.TimerHandle | None = None
         self._custom_attribute_poller_task: asyncio.Task | None = None
         self._attribute_update_callbacks: dict[int, list[Callable]] = {}
-        self._default_fabric_label: str = "Home Assistant"
+        self._default_fabric_label: str | None = None
 
     async def initialize(self) -> None:
         """Initialize the device controller."""
@@ -282,7 +282,7 @@ class MatterDeviceController:
         raise NodeNotExists(f"Node {node_id} does not exist or is not yet interviewed")
 
     @api_command(APICommand.SET_DEFAULT_FABRIC_LABEL)
-    async def set_default_fabric_label(self, label: str) -> None:
+    async def set_default_fabric_label(self, label: str | None) -> None:
         """Set the default fabric label."""
         self._default_fabric_label = label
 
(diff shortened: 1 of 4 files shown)


matter_server/server/device_controller.py
```diff
@@ -315,6 +315,14 @@ async def commission_with_code(
             LOGGER.info("Commissioned Node ID: %s vs %s", commissioned_node_id, node_id)
             if commissioned_node_id != node_id:
                 raise RuntimeError("Returned Node ID must match requested Node ID")
+
+            await self._chip_device_controller.send_command(
+                node_id,
+                0,
+                Clusters.OperationalCredentials.Commands.UpdateFabricLabel(
+                    "Home Assistant"
```
marcelveldt, line 323: I think we should provide this value when starting the server

Apollon77: Apple sets the "Home Name" as label, so yes maybe something the user can configure in settings?

agners: In Home Assistant we have a instance name (aka. location_name, which is also used in DNS-SD as instance name). I've intended to use that:

image

marcelveldt (approved):
Nice addition!

matter_server/server/device_controller.py
```diff
@@ -166,6 +166,7 @@ def __init__(
         self._custom_attribute_poller_timer: asyncio.TimerHandle | None = None
         self._custom_attribute_poller_task: asyncio.Task | None = None
         self._attribute_update_callbacks: dict[int, list[Callable]] = {}
+        self._default_fabric_label: str = "Home Assistant"
```
MartinHjelmare, line 169: I think a generic name eg the library name is more appropriate. Home Assistant can change this as needed.

MartinHjelmare (changes requested):

matter_server/server/device_controller.py
agners: Yeah, I had "Python Matter Server" first, but then reverted to this.

The reason is that if we release this as a non-breaking change, then folks who have an older Core version will commission devices with that default label.

So we have two options here:
a) Make it a non-breaking change and accept the default fabric label (whichever we are going for)
b) Declare this as a breaking change, bump the minimum Core version in the add-on once we release the first Matter Server with this in.

Maybe a third option is to have no default at all? :thinking: 

c) Declare None as default and not set a label in this case?

MartinHjelmare: No label as default sounds ok.

agners: Changed accordingly.

MartinHjelmare (approved):

PR #909, opened 2024-10-01
Update label on interview

agners:
When triggering an interview update the Fabric label on the device as well. This allows users to update the label by a simple re-interview.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -345,7 +345,7 @@ class MatterDeviceController:
         retries = 3
         while retries:
             try:
-                await self.interview_node(node_id)
+                await self._interview_node(node_id)
             except (NodeNotResolving, NodeInterviewFailed) as err:
                 if retries <= 0:
                     raise err
@@ -425,7 +425,7 @@ class MatterDeviceController:
         retries = 3
         while retries:
             try:
-                await self.interview_node(node_id)
+                await self._interview_node(node_id)
             except NodeInterviewFailed as err:
                 if retries <= 0:
                     raise err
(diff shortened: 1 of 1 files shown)


agners:
> Hmm, not sure if I like this side effect. It feels a tiny bit strange to do this in the client library from a purist perspective.

Yeah similar, it feels a bit "wrong" in a way. But then, I felt reinterview is soemthing which doesn't happen often.

Currently it would happen in three cases:
a) When the user asks for a reinterview (the case I aim for)
b) When there was a software update (which triggers a reinterview)
c) When there was a interview schema change

> At the same time it does make sense to do this in one sweep. Maybe we should perform this action from HA ?

Well, I intended more that the user has a way to actually update the label currently (via reinterview command).

Setting the labels on all existing commissioned devices is not the aim here.

marcelveldt:
Yes, lets only do it in case of a) 
We can then argue to do the implementation server side or client side.
Let's keep it simple now and do it client side.

marcelveldt (approved):

hoppel118:
Hi,

do I understand it right, that we now should see a label instead of 4939, after re-interviewing Matter devices?

If so, it doesn’t work in my case. I successfully re-interviewed some of my devices in Home Assistant, but the fabric label in Apple Home is still 4939:

IMG_0641
IMG_0640

The name of my Home Assistant instance is ‘Home Assistant‘:

IMG_0642

My understanding is that ‘Home Assistant‘ should be the new fabric label in Apple Home. Is this right?

Everything is up-to-date:

Core 2024.10.1
Supervisor 2024.10.0
Operating System 13.1
Frontend 20241002.2
HA Matter Server 6.6.0

Thanks Hoppel

agners:
> My understanding is that ‘Home Assistant‘ should be the new fabric label in Apple Home. Is this right?

What Apple prints under Connected Services are two things: The vendor and the fabric label. The vendor name (printed in the larger font) comes from the DCL and recently got set to "Nabu Casa" and then we changed it to "Home Assistant (Nabu Casa)" (see the [Nabu Casa 4939 DCL entry](https://on.dcl.csa-iot.org/dcl/vendorinfo/vendors/4939)). In my Apple Home it currently shows as "Nabu Casa", but I expect that Apple will fetch the updated vendor information over time.

The Fabric label will be set on commissioning or interview. It will be set to the Home Assistant instance name (as set in Settings > System-> General, also known as location name). The [PR on Core side](https://github.com/home-assistant/core/pull/127252) got merged after the last release, so it will only be part of Home Assistant 2024.11.0 (or current dev versions).

hoppel118:
@agners Great, thanks for the detailed information.

Can’t wait for the day when Home Assistant got certified. 😃

Thanks again for all the work the Home Assistant Team investigates into Matter, especially @marcelveldt and you.

Regards Hoppel

PR #970, opened 2024-11-25
(fix/doc) Move the Python example a little further in README

lboue:
Move the Python example a little further in README


marcelveldt (approved):
Thanks!

PR #974, opened 2024-11-26
Revert bump to Matter 1.4 wheels and device types

marcelveldt:
The Matter 1.4 wheels are built against a glibc version that is too new.
We need to revert the bump to 1.4 for now until we resolve that.


agners (approved):

nerijus:
But version 6.6.1 already has Debian 12?

Issue #977, opened 2024-11-27
CHIP Persistent data storage writes are not power-loss safe


kepstin:
In this discord thread: https://discord.com/channels/330944238910963714/1311124568081174628 a user reported an issue where following a power loss, all of their matter devices disappeared. From the Matter server logs:

```
2024-11-26 20:12:41.236 (MainThread) ERROR [chip.storage] Expecting ',' delimiter: line 29 column 111 (char 4111)
2024-11-26 20:12:41.236 (MainThread) CRITICAL [chip.storage] Could not load configuration from /data/chip.json - resetting configuration...
```

It looks like the configuration file was present, but had invalid data in it. The user was able to restore a backup correct the issue, but when I looked at the code to persist the storage: (edit: this code doesn't seem to be responsible for the user's issue, which happened in the Matter SDK code - but this code *is* incorrect.)

https://github.com/home-assistant-libs/python-matter-server/blob/a55bb8b0dc838cfee2eb0dd70f18655e26540efb/matter_server/server/storage.py#L149-L163

I noticed that the logic here is not power loss safe, and might result in a partially written or missing file for unlucky users.

The current logic does:
* Rename `chip.json` to `chip.json.backup`. Following this, if power is lost, there may be no `chip.json` file present at all.
* Write a new `chip.json` file. If the write is interrupted, or power is lost before the system flushes the write cache, a partially written or corrupt configuration file may be present on reboot, resulting in errors like this user saw.

I'd recommend the following logic instead:
* (optional) Copy `chip.json` to `chip.json.backup`, leaving the existing file in place. (Ideally, run fsync on the newly written file to ensure the backup is updated before the original file gets replaced.)
* Write the new data to a temporary file, like `chip.json.new`.
* Run fsync (or fdatasync) on the newly written file to ensure the file contents are persisted on the disk. (see the notes on https://docs.python.org/3/library/os.html#os.fsync regarding flushing the python buffers). This is needed to ensure ordering - making sure that the file contents will be written *before* the file gets renamed.
* Rename the temporary file to the final filename, atomically replacing the existing file.
* (optional) Run fsync on the renamed file to ensure the rename is committed to storage. (This is only needed if data stored elsewhere - e.g. in a sqlite database - might depend on this data having been persisted)

kepstin:
… Actually I think I'm looking at the wrong code here - the error messages don't match up. Is the code in `matter-server/server/storage.py` used? is it for something else?

[...]

PR #1014, opened 2024-12-23
Use atomic writes for the storage file

marcelveldt:
Prevent our persistent storage file get corrupted if writing fails or power is cut during writing.
This is step 1 because a similar approach is also needed for the sdk's json file which contains the fabric (cryptographic) data.

Related to #977

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ server = [
   "aiohttp==3.11.10",
   "aiorun==2024.8.1",
   "async-timeout==5.0.1",
+  "atomicwrites==1.4.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -35,6 +35,7 @@ server = [
   "aiohttp==3.11.10",
   "aiorun==2024.8.1",
   "async-timeout==5.0.1",
+  "atomicwrites==1.4.0",
```
agners, line 38: There is 1.4.1 it seems? :thinking: 

```suggestion
  "atomicwrites==1.4.1",
```

agners (approved):

PR #1022, opened 2024-12-30
Add step 'Activate the new Python environment' in DEVELOPMENT.md

lboue:
Add step 'Activate the new Python environment'

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -6,6 +6,7 @@ Development is only possible on a (recent) Linux or MacOS machine. Other operati
 
 - Download/clone the repo to your local machine.
 - Set-up the development environment: `scripts/setup.sh`
+- Activate the new Python environment  `source .venv/bin/activate`
 
 ## Start Matter server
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -6,6 +6,7 @@ Development is only possible on a (recent) Linux or MacOS machine. Other operati
 
 - Download/clone the repo to your local machine.
 - Set-up the development environment: `scripts/setup.sh`
+- Activate the new Python environment  `source .venv/bin/activate`
```
marcelveldt, line 9: this should be included in the setup script so if that is not the case, we should adjust that

marcelveldt: Yeah, its already part of the setup script

marcelveldt: Basically you should ALWAYS run the setup script to prepare your dev environment

lboue:
Indeed, it should have worked. I'll check it out.

PR #1027, opened 2025-01-09
Support local updates with the same version string

agners:
For development it is sometimes useful to just reuse the same version string but increment the underlying build number (which is the actual version devices typically care about).

This change supports multiple local updates with the same version string, and updating to an explicit version by version integer.

### DIFF
--- a/matter_server/server/ota/__init__.py
+++ b/matter_server/server/ota/__init__.py
@@ -8,7 +8,9 @@ from pathlib import Path
 from matter_server.common.models import UpdateSource
 from matter_server.server.ota import dcl
 
-_local_updates: dict[tuple[int, int], dict] = {}
+MatterProduct = tuple[int, int]
+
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):

matter_server/server/ota/__init__.py
```diff
@@ -38,14 +43,16 @@ async def check_for_update(
...
-            requested_software_version is None
-            or local_update["softwareVersion"] == requested_software_version
-            or local_update["softwareVersionString"] == requested_software_version
-        ):
-            return UpdateSource.LOCAL, local_update
+    if local_updates := _local_updates.get((vid, pid)):
+        logger.info("Local updates found for this device")
+        if requested_software_version is None:
+            # Use integer version to reliably determine absolute latest version
+            versions = filter(
```
MartinHjelmare, line 50: A list comprehension is easier to read and faster.

PR #1081, opened 2025-02-19
Developer documentation: /data & IPv4 group limits

markus-becker-tridonic-com:
Fixes https://github.com/home-assistant-libs/python-matter-server/issues/1075.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -6,6 +6,7 @@ Development is only possible on a (recent) Linux or MacOS machine. Other operati
 
 - Download/clone the repo to your local machine.
 - Set-up the development environment: `scripts/setup.sh`
+- Create the `/data` directory if it does not exist with permissions for the user running the python-matter-server.
 
 ## Start Matter server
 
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
Thanks Markus!

DEVELOPMENT.md
```diff
@@ -19,6 +20,12 @@ Use `--help` to get a list of possible log levels and other command line argumen
 
 The server runs a Matter Controller and includes all logic for storing node information, interviews and subscriptions. To interact with this controller we've created a small Websockets API with an RPC-like interface. The library contains a client as reference implementation which in turn is used by Home Assistant. Splitting the server from the client allows the scenario where multiple consumers can communicate to the same Matter fabric and the Matter fabric can keep running while the consumer (e.g. Home Assistant is down).
 
+If you happen to get `OSError: [Errno 105] No buffer space available.`, increase the IPv4 group limits with:
+```
+echo "net.ipv4.igmp_max_memberships=1024" >> sysctl.conf
+sudo sysctl -p sysctl.conf
```
zackelia, line 26: This solution is not permanent and will be reset back to its default value on reboot. What about a solution like this following the recommendations from `/etc/sysctl.d/README.sysctl`?
```
echo "net.ipv4.igmp_max_memberships=1024" | sudo tee -a /etc/sysctl.d/local.conf
sudo service procps force-reload
```

zackelia, line 26: This solution is not permanent and will be reset back to its default value on reboot. What about a solution like this following the recommendations from `/etc/sysctl.d/README.sysctl`?

```suggestion
echo "net.ipv4.igmp_max_memberships=1024" | sudo tee -a /etc/sysctl.d/local.conf
sudo service procps force-reload
```

agners (approved):

PR #1120, opened 2025-04-03
Improve mDNS browsing to find operational nodes

agners:
Use the fabric specific DNS-SD service subtype instead of the generic "_matter._tcp" service type. This avoids querying for operational nodes of other fabrics. This is closer to the recommendations in "4.3.2.6. Performance Recommendations" of the Matter spec.

While at it, remove the commissionable node browsing as it is not used currently. This too lowers the unnecessary mDNS traffic on the network.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -24,7 +24,12 @@ from chip.clusters.ClusterObjects import ALL_ATTRIBUTES, ALL_CLUSTERS, Cluster
 from chip.discovery import DiscoveryType
 from chip.exceptions import ChipStackError
 from chip.native import PyChipError
-from zeroconf import BadTypeInNameException, IPVersion, ServiceStateChange, Zeroconf
+from zeroconf import (
+    DNSQuestionType,
+    IPVersion,
+    ServiceStateChange,
+    Zeroconf,
+)
 from zeroconf.asyncio import AsyncServiceBrowser, AsyncServiceInfo, AsyncZeroconf
 
 from matter_server.common.const import VERBOSE_LOG_LEVEL
@@ -215,11 +220,14 @@ class MatterDeviceController:
         LOGGER.info("Loaded %s nodes from stored configuration", len(self._nodes))
         # set-up mdns browser
         self._aiozc = AsyncZeroconf(ip_version=IPVersion.All)
-        services = [MDNS_TYPE_OPERATIONAL_NODE, MDNS_TYPE_COMMISSIONABLE_NODE]
+        services = [
+            f"_I{self.compressed_fabric_id:0{16}X}._sub.{MDNS_TYPE_OPERATIONAL_NODE}"
+        ]
         self._aiobrowser = AsyncServiceBrowser(
             self._aiozc.zeroconf,
             services,
             handlers=[self._on_mdns_service_state_change],
+            question_type=DNSQuestionType.QM,
         )
 
     async def stop(self) -> None:
@@ -1510,13 +1518,6 @@ class MatterDeviceController:
             # We already have a timer to resolve this service, so ignore this callback.
             return
 
-        if service_type == MDNS_TYPE_COMMISSIONABLE_NODE:
-            # process the event with a debounce timer
-            self._mdns_event_timer[name] = self._loop.call_later(
-                0.5, self._on_mdns_commissionable_node_state, name, state_change
-            )
-            return
-
         if service_type == MDNS_TYPE_OPERATIONAL_NODE:
             if self._fabric_id_hex is None or self._fabric_id_hex not in name.lower():
                 # filter out messages that are not for our fabric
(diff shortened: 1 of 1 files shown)


agners:
There is a slight chance that some misbehaving devices do not announce themselfs using the subtype. In my testing with Thread border routers from Apple, Google and our OTBR Add-on as well as all the WiFi devices I have on my home things seem to work as usual. Since Python Zeroconf only returns the matching devices, and we rely on the browsing result to connect to devices in first place, verifying subtype DNS-SD works for all devices in a home is rather easy: Run the Python Matter server with this code and see if all devices come online.

Apollon77:
@agners The other question is why you query MDNS at all without a need? ;-)

Ideally you should try to connect to the last known address of commissioned nodes and do not need to query anything at all. And when the last known address does not work then you discover this device specifically. Thats at least how I understood it to ideally work on controller side.

marcelveldt:
> @agners The other question is why you query MDNS at all without a need? ;-)
> 
> Ideally you should try to connect to the last known address of commissioned nodes and do not need to query anything at all. And when the last known address does not work then you discover this device specifically. Thats at least how I understood it to ideally work on controller side.

Its for fast operational node discovery.
Imagine somebody removing the actual power from a lightbulb. Then when power is re-applied, the expectation is that the bulb is rediscovered again (so working) within seconds.

By just only try to connect to a device at some interval you will not be able to give that user experience.
So its a combination of trying to connect at X interval and doing the operational node discovery on mdns.

marcelveldt:
> While at it, remove the commissionable node browsing as it is not used currently.

It is being used by Home Assistant to auto discover the presence of a matter device. It should not be removed.

matter_server/server/device_controller.py
```diff
@@ -1510,13 +1518,6 @@ def _on_mdns_service_state_change(
             # We already have a timer to resolve this service, so ignore this callback.
             return
 
-        if service_type == MDNS_TYPE_COMMISSIONABLE_NODE:
-            # process the event with a debounce timer
-            self._mdns_event_timer[name] = self._loop.call_later(
-                0.5, self._on_mdns_commissionable_node_state, name, state_change
-            )
-            return
-
```
marcelveldt, line 1519: Can this be handled in a seperate dedicated PR with a breaking change note ?
Because HA is not the only lib consumer of Matter Server. My preference would be to leave it in and not cause breaking changes.

agners:
> Ideally you should try to connect to the last known address of commissioned nodes and do not need to query anything at all. And when the last known address does not work then you discover this device specifically. Thats at least how I understood it to ideally work on controller side.

Besides the address some of the protocol parameters are part of the TXT section (`SII`/`SAI` parameters). Of course we could cache those as well, and just opportunistically connect to the last known address, but maybe it is just easier to check mDNS to make sure to get the latest? :thinking: :shrug: 

TBH, I think using the SDK asking it to connect to a IP directly is actually not possible. I think we have to create a device controller with the fabric settings, and then ask it to establish a session to a particular node ID. The SDK then requests mDNS (with its natiive C++ minimal mDNS implementation)  to get the IP and handles all that.

You might now ask why do we use another mDNS in the Python Matter Server. The main reason is that the SDK doesn't have a mechanism to listen continuously for devices coming online. In theory we only need that part to be implemented on the Python side. I am actually looking into using Python Zeroconf to only passively listen for devices coming back online and leave all the rest to the SDK.

matter_server/server/device_controller.py
agners: > Can this be handled in a seperate dedicated PR with a breaking change note ?

Removed in #1121.

> Because HA is not the only lib consumer of Matter Server. My preference would be to leave it in and not cause breaking changes.

I don't think this influences our server API in any way. We can simply revert the above PR if we actually make use of it.

marcelveldt:
> The main reason is that the SDK doesn't have a mechanism to listen continuously for devices coming online.

There have been several conversations about operational node discovery within the SDK. Did none of that get implemented in the meanwhile ?

3oris:
@marcelveldt the openthread people seem to be working on this if I am not mistaken. 

There seems to be a major overhaul of mdns/dnssd/srp stuff ongoing losely tracked here
https://github.com/openthread/openthread/issues/9434 which ultimatly should also yield a standalone mdns implementation, because

> However, there is interest in allowing this mDNS implementation to be used and integrated into other projects without including the full Thread stack.

See here: https://github.com/openthread/openthread/pull/11191 and here: https://github.com/openthread/openthread/blob/15553e6dd22368204342844ee101a4b7fc2c0b00/examples/platforms/posix_mdns/README.md

Not sure, how much of this can actually be used other contexts though.

marcelveldt:
Yes I'm aware of that but we're also talking about general Matter operational node discovery, not only Thread. 

Let's hope we can get to one general mdns implementation because the current situation is just hopeless with every implementation having their own quirks

Apollon77:
@marcelveldt Re your answer above: Ok if that's mainly about listening and not active querying then all makes sense :-)

agners:
Part of this PR is in #1121, other parts are in #1127. The rest would need a new PR, but there is no urgency for this right now.

PR #1121, opened 2025-04-03
Remove commissionable node browsing

agners:
Remove commissionable node browsing from the device controller. It is currently not used and causes additional mDNS traffic.

### DIFF
--- a/matter_server/server/device_controller.py
+++ b/matter_server/server/device_controller.py
@@ -24,7 +24,7 @@ from chip.clusters.ClusterObjects import ALL_ATTRIBUTES, ALL_CLUSTERS, Cluster
 from chip.discovery import DiscoveryType
 from chip.exceptions import ChipStackError
 from chip.native import PyChipError
-from zeroconf import BadTypeInNameException, IPVersion, ServiceStateChange, Zeroconf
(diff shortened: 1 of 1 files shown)


marcelveldt (changes requested):
We have no idea if this is not in use with any of the lib consumers. There are other projects nowadays depending on our server. Is it really needed to remove this ? I can't imagine this causes that much traffic.

If it is, we need to deprecate it first by logging a warning for a couple of months and then remove it.

3oris:
AND/OR you introduce a feature toggle with variant

1. yielding the deprecation warning alongside the old behaviour
2. disabling the code in question to reduce mdns traffic for the Home Asisstant use case

Having that toggle in place might also facilitate measuring the actual mdns traffic delta in an otherwise identical setup

agners:
> Having that toggle in place might also facilitate measuring the actual mdns traffic delta in an otherwise identical setup

Yeah I was considering something like that, e.g. we could only start active browsing when `start_listening` command is sent, potentially with an argument.

However, it seems that browsing is actually switching to a passive mode after 4 tries. Having a continuous browse doesn't seem as bad as I initially thought. Given that, and the fact that we might have users using this, let's close this.

PR #1137, opened 2025-05-02
Add Mounted device types (1.4)

lboue:
Adding the new mounted device types (1.4).

### Testing

image

### DIFF
--- a/matter_server/client/models/device_types.py
+++ b/matter_server/client/models/device_types.py
@@ -154,6 +154,30 @@ class DimmableLight(DeviceType, device_type=0x0101):
     }
 
 
+class MountedOnOffControl(DeviceType, device_type=0x010F):
+    """Mounted On/Off Control."""
+
+    clusters = {
+        all_clusters.Groups,
+        all_clusters.Identify,
+        all_clusters.LevelControl,
+        all_clusters.OnOff,
+        all_clusters.ScenesManagement,
+    }
+
+
+class MountedDimmableLoadControl(DeviceType, device_type=0x0110):
+    """Mounted Dimmable Load Control."""
+
+    clusters = {
+        all_clusters.Groups,
+        all_clusters.Identify,
+        all_clusters.LevelControl,
+        all_clusters.OnOff,
+        all_clusters.ScenesManagement,
+    }
+
+
 class ColorTemperatureLight(DeviceType, device_type=0x010C):
     """Color Temperature Light."""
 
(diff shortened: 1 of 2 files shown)


matter_server/client/models/device_types.py
```diff
@@ -154,6 +154,30 @@ class DimmableLight(DeviceType, device_type=0x0101):
     }
 
 
+class MountedOnOffControl(DeviceType, device_type=0x010F):
+    """Mounted On/Off Control."""
+
+    clusters = {
```
agners, line 160: At least for other device types like `DimmableLight` we have the Base Device Type clusters here too (`Descriptor` and `Binding`).

Also, `LevelControl` is an optional cluster, and so is `OccupancySensing`. I think we typically add optional clusters here.

agners (changes requested):

lboue:
> Normally we auto generate this with the script

I've posted it in the SDK repo. It was an oversight on their part.
There is a PR to fix that:
- Issue https://github.com/project-chip/connectedhomeip/issues/38718
- PR: https://github.com/project-chip/connectedhomeip/pull/38737

agners:
> Normally we auto generate this with the script

Hm, I see the generated output from current master generates like this:

```python
class MountedOnOffControl(DeviceType, device_type=0x010F):
    """Mounted On/Off Control."""

    clusters = {
        all_clusters.Groups,
        all_clusters.Identify,
        all_clusters.LevelControl,
        all_clusters.OnOff,
        all_clusters.ScenesManagement,
    }


class MountedDimmableLoadControl(DeviceType, device_type=0x0110):
    """Mounted Dimmable Load Control."""

    clusters = {
        all_clusters.Groups,
        all_clusters.Identify,
        all_clusters.LevelControl,
        all_clusters.OnOff,
        all_clusters.ScenesManagement,
    }
```

So it is missing these base device type clusters too :shrug: 

I remember that source XML was always buggy, unfortunately. Since I don't think the clusters have real-world impact right now, I am fine adding those as is (since that is how the XML looks like currently too), and then later on fix the upstream XML. What do you think @marcelveldt .

agners:
> I've posted it in the SDK repo. It was an oversight on their part.

Nice! That said, our script uses master branch anyways, see https://github.com/home-assistant-libs/python-matter-server/blob/main/scripts/generate_devices.py#L12-L13  :sweat_smile: In the end, I don't think it is important to have v1.4-branch updated, that XML is not well maintained anyways. IMHO, more important would be to have it correct in master branch.

As for the mounted device types, the base device types and the occupancy sensing are definitly missing. When you look at On/Off light, the optional clusters are there. So this actually needs some updates upstream as well.

agners (approved):
Since this is the current state of the device types XML, let's just go with this.

PR #1144, opened 2025-05-07
Improve Matter DNS-SD service parsing

agners:
Services are now parsed in a more robust way, to avoid issues in cases where the service name is not exactly as expected. This has been observed to be the case for Thread devices, presumably because two Thread border routers attempt to announce the same service name. The service name then becomes a " (2)" suffix, which the old code did not handle well.

We could parse this duplicated messages as well, but generally only a single node should ever announce a service name. If a service name gets announced by two independent devices, the two services likely stem from the same underlying device but simply got duplicated by some environmental factor (e.g. Thread border routers not being in sync on who announces the service name, or badly configured mDNS reflectors duplicating messages on the same network). Only processing the correctly formatted service name is sufficient in these cases.

### DIFF
--- a/matter_server/server/server.py
+++ b/matter_server/server/server.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 from functools import cached_property, partial
+import inspect
 import ipaddress
 import logging
 import os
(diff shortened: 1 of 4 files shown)


agners:
This prevents errors like this:

```
2025-04-24 13:46:58.616 (MainThread) ERROR [matter_server.server] Error doing task: Exception in callback MatterDeviceController._on_mdns_operational_node_state('A3604AC231F8...r._tcp.local.', <ServiceState...ge.Updated: 3>)
  File "/usr/local/lib/python3.12/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/lib/python3.12/site-packages/matter_server/server/device_controller.py", line 1544, in _on_mdns_operational_node_state
    node_id = int(name.split("-")[1].split(".")[0], 16)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 16: '00000000000000AD (6)'
```

marcelveldt (approved):
LGTM!
St some point we should write a simple test to test a bunch of dns names against this regex

matter_server/server/server.py
```diff
@@ -315,18 +316,24 @@ def register_api_command(
...
+                if not hasattr(attr, "api_cmd"):
                     continue
-                if hasattr(val, "mock_calls"):
-                    # filter out mocks
+                if hasattr(attr, "mock_calls"):
                     continue
-                # method is decorated with our api decorator
-                self.register_api_command(val.api_cmd, val)
+
+                # Bind the function to the instance before registering
```
MartinHjelmare, line 334:
```suggestion
                # Get the instance method before registering
```

PR #1155, opened 2025-05-19
add node binding dialog.

Luxni:
I added a button in the binding cluster of the node to add the binding relationship of the device.
Note: Currently, binding to the endpoints of the device is only supported.

I conducted tests using [light_switch](https://github.com/espressif/esp-matter/tree/main/examples/light_switch) and [light](https://github.com/espressif/esp-matter/tree/main/examples/light) in the esp-matter sdk, and the test results met the requirements.
(One-to-one endpoint control can be achieved, as well as one-to-many endpoint control.)

The operation is as follows
1. Enter the binding cluster of the node, and a 'Binding' button will appear on it.
2. After clicking the 'Binding' button, you can add or delete binding relationships.

As shown in the following figure:

### DIFF
--- a/dashboard/note.md
+++ b/dashboard/note.md
@@ -1,0 +1,12 @@
+```
+                          <div
+                            style="display:flex;flex-direction:row;gap:15px;"
+                          >
+                            <div>${"privilege:" + entry.privilege}</div>
+                            <div>${"authMode:" + entry.authMode}</div>
+                            <div>${"fabricIndex:" + entry.fabricIndex}</div>
+                          </div>
+                          <div>subjects:${JSON.stringify(entry.subjects)}</div>
+                          <div>targets:${JSON.stringify(entry.targets)}</div>
+
+```
--- a/dashboard/src/client/client.ts
+++ b/dashboard/src/client/client.ts
@@ -143,6 +143,21 @@ export class MatterClient {
     await this.sendCommand("update_node", 10, { node_id: nodeId, software_version: softwareVersion });
   }
 
+  async setACLEntry(nodeId: number, entry: any) {
+    return await this.sendCommand("set_acl_entry", 0, {
+      node_id: nodeId,
+      entry: entry,
+    });
+  }
+
+  async setNodeBinding(nodeId: number, endpoint: number, bindings: any) {
+    return await this.sendCommand("set_node_binding", 0, {
+      node_id: nodeId,
+      endpoint: endpoint,
+      bindings: bindings,
+    });
+  }
+
   async sendCommand<T extends keyof APICommands>(
     command: T,
     require_schema: number | undefined = undefined,
--- a/dashboard/src/client/models/model.ts
+++ b/dashboard/src/client/models/model.ts
@@ -105,6 +105,14 @@ export interface APICommands {
     requestArgs: {};
     response: {};
   };
+  set_acl_entry: {
+    requestArgs: {};
+    response: {};
+  };
+  set_node_binding: {
+    requestArgs: {};
+    response: {};
+  };
 }
 
 export interface CommandMessage {
--- a/dashboard/src/components/dialogs/acl/model.ts
+++ b/dashboard/src/components/dialogs/acl/model.ts
@@ -1,0 +1,109 @@
+
+export type AccessControlEntryRawInput = {
+  "1": number;
+  "2": number;
+  "3": number[];
+  "4": null;
+  "254": number;
+};
+
+export type AccessControlTargetStruct = {
+  cluster: number | undefined;
+  endpoint: number | undefined;
+  deviceType: number | undefined;
+};
+
+export type AccessControlEntryStruct = {
+  privilege: number;
+  authMode: number;
+  subjects: number[];
+  targets: AccessControlTargetStruct[] | undefined;
+  fabricIndex: number;
+};
+
+export class AccessControlTargetTransformer {
+  private static readonly KEY_MAPPING: {
+    [inputKey: string]: keyof AccessControlTargetStruct;
+  } = {
+    "0": "cluster",
+    "1": "endpoint",
+    "2": "deviceType",
+  };
+
+  public static transform(input: any): AccessControlTargetStruct {
+    if (!input || typeof input !== "object") {
+      throw new Error("Invalid input: expected an object");
+    }
+
+    const result: Partial<AccessControlTargetStruct> = {};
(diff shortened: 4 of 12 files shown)


Apollon77:
Cool idea,

formally the ACL should limit access to one cluster of the endpoint and it should also made sure that also a cluster client is defined for that on the target. Could that get added?

Luxni:
@Apollon77 

Is it as shown in the following figure? 

Screenshot_2025-05-20_22-20-23


command like this :
```typescript
const bindingEntry: BindingEntryStruct = {
    node: targetNodeId,
    endpoint: targetEndpoint,
    group: undefined,
    cluster: targetCluster,
    fabricIndex: undefined,
};
```

lboue:
> Cool idea,
> 
> formally the ACL should limit access to one cluster of the endpoint and it should also made sure that also a cluster client is defined for that on the target. Could that get added?

I think that in some cases like dimmer switch this may require pairing with several clusters, such as cluster no. 6 is the On/Off cluster and the cluster no. 8 is the Level Control cluster. 
See this:
https://github.com/project-chip/connectedhomeip/tree/master/examples/light-switch-app/nrfconnect#unicast-binding-to-a-remote-endpoint-using-the-chip-tool-for-windows-or-linux

```
In this command:

{"fabricIndex": 1, "node": <1>, "endpoint": 1, "cluster": 6} is a binding for the On/Off cluster.
{"fabricIndex": 1, "node": <1>, "endpoint": 1, "cluster": 8} is a binding for the Level Control cluster.
```

Apollon77:
Yes formally it should be mapped to "where a cluster client existing for" ... so yes having the acl for the endpoint is pragmatic solution :-)

lboue:
I ran a test with the latest commit and it worked fine. I tried with 2 DK:
  nRF5340 DK as a light
  SL xG24 Explorer Kit as a switch

image

A-vesalius:
ran a test between Inovelli vtm31 white dimmer switch and aqara T2 as well Nanoleaf essentials NL67 bulbs Dimming & On/Off works from the switch paddle. Virtually controlling the Inovelli switch (on/off/dim) thru HA or Apple Home does not control bulbs though. 
Screenshot 2025-06-23 at 11 02 46 PM
Screenshot 2025-06-23 at 11 04 09 PM
Screenshot 2025-06-23 at 11 03 50 PM

Apollon77:
Cool, but be careful with the ACLs ... number of ACLs per fabric is limited. so if you needed to setup two ACLs here you already use a lot of slows on the node for the fabric. That's one of the fun parts around bindings or groups ...

A-vesalius:
@Apollon77 thanks deleted 3 of the 4 bindings to constrain variables.

@Luxni ACL written to the target device successfully. Will check the debug panel when off of work. 

to be clear neither bulb (aqara/Nanoleaf) is physically connected to the and separately powered from a wall outlet. The vtm31 is controlling both using bindings when the switch paddle is physically pressed. The switch controls neither when dim or on/off controlled virtually via HA or apple home.

Apollon77:
@harryh What you mean by "binding tow dimmer switches together"? What exactly you want to achieve? Normally bindings wrong from switch to "light bulb" (in that case)

A-vesalius:
@araemo 
since reporting this early on, we have learned this is expected behavior with matter bindings as they are setup in the current matter spec, not just an inovelli issue. Neither intuitive nor optimal at the moment. see the discord discussion below.

https://discord.com/channels/330944238910963714/1368077978319327323/1397949338876903545

Apollon77:
@araemo 
> Re "pressing the on/off or dim control via HA, the switch does not send messages to the bound target. "

I think you confuse features. The python matter server acts as a controller. So when there is a device that can be controlled e.g. onoff then you get such an entity and you control this device. Done!

Binds are a Matter feature for other nodes in the matter system like a light switch(e,g, the innovelli switches) that can directly control "Bound devices" without the need for the controller to do anything. Setting up a binding defines a "direct communication route between matter nodes.

This means that HA with the python matter node never will use the bindings itself. It will allow you to set them up for all devices in the matter fabric.

araemo:
> @araemo
> 
> > Re "pressing the on/off or dim control via HA, the switch does not send messages to the bound target. "
> 
> I think you confuse features. The python matter server acts as a controller. So when there is a device that can be controlled e.g. onoff then you get such an entity and you control this device. Done!
> 
> Binds are a Matter feature for other nodes in the matter system like a light switch(e,g, the innovelli switches) that can directly control "Bound devices" without the need for the controller to do anything. Setting up a binding defines a "direct communication route between matter nodes.
> 
> This means that HA with the python matter node never will use the bindings itself. It will allow you to set them up for all devices in the matter fabric.

To be clear, I am not expecting HA to 'honor' the binding or use it.  I would expect if I turn on the switch, that the light turns on.  Regardless if I turn it on by pressing the physical button or the HA button. (And to be very clear: I would expect the switch to send the message to its bound target, not HA)  I see the comment above that the matter standard explicitly says that shouldn't happen, but that is not what I expected.  (And even though I don't have experience with zigbee bindings, I gather that zigbee bindings DID work that way, since other people with zigbee experience are also surprised by that)

araemo:
So, for example, I have an inovelli VTM-31-SN (their matter/thread dimmer switch), and bound it to an inovelli VTM-36 (the lightfan module, the switch is bound to the light endpoint, not the fan endpoint).

The two things that go 'wrong' if I control the light directly from HA/other fabrics, instead of controlling both the light and the switch:
1. The switch has an LED bar to indicate the light state and brightness level.  If I send a command to the light directly, this LED bar is not updated with the light's new state.
2. If I use HA/another fabric to turn the light off, but do not also send an 'off' message to the switch, the switch's on button doesn't turn the light on.  Probably related to 1, since it seems to be based on the switch's tracking of the light's state.

If these two facts are due to how matter/thread is supposed to work, I would think instead of adding the light itself as your controllable entity, you'd want to add the switch as the controllable entity, and let the binding make sure they stay synced.

But since that doesn't work, it is just a bad experience.  I've grouped the switch and the light in the same room in other fabrics, but since my main control is voice control from google home, I just put both in the same room and google home tells me 'two lights on' or 'two lights off' when the 'two lights' are the switch and the light.

I do think it would be more logical for the light to be able to inform the switch of state changes, and have the switch update its LED and internal state tracking to match, instead of getting out of sync and no longer controlling the light properly.  But as it stands now, it's just a poor user experience.  (And to be clear, I'm not blaming HA for that, I'm blaming inovelli and/or the matter standard)

Issue #1198, opened 2025-07-20
Deleted Bindings also need to delete ACLs


sup3rm3ga:
I successfully added bindings between my Inovelli White switch and Nanoleaf lights. I deleted a few of the bindings and tried to re-add them, but it seems every deleted binding cannot be re-added. I'm going to factory reset the lights and try again.

jimmyobomsawin:
I think I'm having the same problem. At first I couldn't figure out the pattern as to why I couldn't bind lights. Once I saw your post I realized that that's exactly what it was - if I added a light to my binding and then removed it, I could no longer add the light to any of my Inovelli White switches. And it seems to block the entire node. I added a Philips Hue light to a switch to test it out and then removed it. Now I am no longer able to add any Hue lights, which all share the same node. When I try to add, the UI just doesn't respond. No error message, nothing. The "add" button is just dead.

sup3rm3ga:
I'm not too sure about Matter or how this code works, but I'm wondering if it is writting the access control list or something on the light node but that isn't getting cleaned up, so when it tries to process it again it errors out for some reason.

Apollon77:
Whats the error when you try? Could it be that you run into ACL limits? Any chance to get a log or such?

Luxni:
I'm very sorry. As I'm currently busy with another project, I only noticed this issue now.

I got the question on discord.
Now I need some information to reproduce this problem.

Please follow the instructions below:
1. Open the developer tools in your browser.
2. Open the binding interface and fill in the information.
3. Clear the console information. There is a button in the upper left corner of the console.
4. Press the binding button, view the information generated in the console。
5. take a screenshot to send here.

jvmahon:
@Luxni 
Here's the experiment. I'm binding one Inovelli VTM31 to another
In this case, I'm binding node 186 to node 188.


Steps 1:
First, I bound node 186 to endpoint 1, cluster 6
Second, I bound node 186 to endpoint 6, cluster 6

Then I deleted both those bindings.

Then I checked the ACL on Node 188.

<img width="720" height="283" alt="Image" src="https://github.com/user-attachments/assets/487ddcb3-15ed-4034-8f33-fffb6a49e109" />

Or reformatted as a JSON structure, below.
As you can see, the entries for the bindings are still in the ACL after the deletion.

`````
[
  {
    "1": 5,
    "2": 2,
    "3": [
      112233
    ],
    "4": null,
    "254": 3
  },
  {
    "1": 5,
[...]

PR #1232, opened 2025-09-01
fix matter binding detele error

Luxni:
repair the binding delete failure problem, caused by error incoming parameters.

### DIFF
--- a/dashboard/src/components/dialogs/acl/model.ts
+++ b/dashboard/src/components/dialogs/acl/model.ts
@@ -1,16 +1,16 @@
 
-export type AccessControlEntryRawInput = {
-  "1": number;
-  "2": number;
-  "3": number[];
-  "4": null;
-  "254": number;
-};
-
 export type AccessControlTargetStruct = {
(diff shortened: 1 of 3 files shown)


Luxni:
#### related issues
1. https://github.com/matter-js/python-matter-server/issues/1198

marcelveldt (approved):
LGTM
Thanks, @Luxni

harryh:
While messing with the new binding features after the first release, I managed to get the ACL on one of my devices in a bad state. It would be nice if there was some way to fix this from the UI.

Screenshot 2025-09-29 at 1 43 04 PM

I know I could just reset and recommission the device, or manually use chip-tool to fix it, but it seems like fixing this from within HA would be helpful.

Luxni:
@harryh 
The project has entered the maintenance stage.
ref: [Announce Matter Server rewrite and maintenance mode](https://github.com/matter-js/python-matter-server/commit/2a4b957357ed7ffcf846ca7b289c0e8b776e4363)

Issue #1182, opened 2025-06-30
Bindings - Specifying Cluster Should be Optional


jvmahon:
@Luxni 

First, let me say that I love this new binding feature and thanks for the work implementing it.

Now, a minor issue ... in the current beta Python Matter server 8.1.0 Beta 0, three parameters (node ID, Endpoint ID, and Cluster ID) must currently be specified to create a binding.

Per the Matter Core Specification Version 1.4.1, Section 9.6.5.1, I believe that the "Cluster ID" is an optional parameter and, if omitted, the binding would be to all eligible clusters on the endpoint. 

Note that the example in Section 9.6.6.1 of the Matter core Spec. shows omission of cluster for a "group" target, but based on the specs., I think the cluster is also omit-able for a non-group node / endpoint target.

Not 100% sure I'm correct here, but flagging this so it can be checked.

Thanks.

Apollon77:
In fact not fully correct. :-)

Here is how I read the spec:

Group and Endpoint fields are mutually exclusive, because when you define a group than the group definition already incudes the endpoint , so that's "derived from the group".  So either Group OR endpoint are set. The nodeid is needed when the endpoint is set. So far so good.

Cluster is Optional by definition. When it is set then it needs to match with an existing client cluster on that list (Description cluster "ClientList"):

> A binding SHALL only be created with the Cluster field if the indicated client cluster exists on the endpoint.

The specification underspecifies the case "cluster field not set" but the one example in 9.6.6.1. Binding Attribute lets assume that omitting the cluster field allows the  binding source to control "all clusters in the clientList of that endpoint". And this would also make sense and be consistent.

So for a fully correct implementation (lets ignore groups for now):
* Read "clientList" of descriptor cluster of the target endpoint
* Allow user to select one of these clusters as "cluster" for the binding to generate and have one option "all" which is simply omitting the field to generate a "all these clusters" binding.

The main problem with bindings these days is that noone knows what exactly do the "Client devices" really support in doing, unless they documented it.

jvmahon:
> Here is how I read the spec:

I'm reading the spec. the same way, though I'd say the spec has multiple ways of being read! In the end, I suppose the real "definitive" source is what the current public SDK (1.4.2) actually implements.

Apollon77:
[...]

PR #1242, opened 2025-09-13
Set the 'cluster' item in 'Binding' as an option

Luxni:
xdph_screenshot_652052b4

### Change
1. determine whether the data entered is correct.
2. prompt whether the input item is necessary or optional.

### related
https://github.com/matter-js/python-matter-server/issues/1182

### DIFF
--- a/dashboard/src/components/dialogs/binding/node-binding-dialog.ts
+++ b/dashboard/src/components/dialogs/binding/node-binding-dialog.ts
@@ -38,13 +38,13 @@ export class NodeBindingDialog extends LitElement {
   @property({ attribute: false })
   endpoint!: number;
 
-  @query("md-outlined-text-field[label='node id']")
+  @query("md-outlined-text-field[name='NodeId']")
   private _targetNodeId!: MdOutlinedTextField;
 
-  @query("md-outlined-text-field[label='endpoint']")
+  @query("md-outlined-text-field[name='Endpoint']")
   private _targetEndpoint!: MdOutlinedTextField;
 
-  @query("md-outlined-text-field[label='cluster']")
+  @query("md-outlined-text-field[name='Cluster']")
   private _targetCluster!: MdOutlinedTextField;
 
   private fetchBindingEntry(): BindingEntryStruct[] {
@@ -264,6 +264,23 @@ export class NodeBindingDialog extends LitElement {
     this.parentNode!.removeChild(this);
   }
 
+  private onChange(e: Event) {
+    const textfield = e.target as MdOutlinedTextField;
+    const value = parseInt(textfield.value, 10);
+
+    if (
+      parseInt(textfield.max, 10) < value ||
+      value < parseInt(textfield.min, 10)
+    ) {
+      textfield.error = true;
+      textfield.errorText = "value error";
+    } else {
+      textfield.error = false;
+    }
+
+    // console.log(`value: ${value} error: ${textfield.error}`);
+  }
+
   protected render() {
     const bindings = Object.values(
       this.node!.attributes[this.endpoint + "/30/0"],
@@ -298,15 +315,33 @@ export class NodeBindingDialog extends LitElement {
               <div class="group-label">target</div>
               <md-outlined-text-field
                 label="node id"
+                name="NodeId"
+                type="number"
+                min="0"
+                max="65535"
                 class="target-item"
+                @change=${this.onChange}
+                supporting-text="required"
               ></md-outlined-text-field>
               <md-outlined-text-field
                 label="endpoint"
+                name="Endpoint"
+                type="number"
+                min="0"
+                max="65535"
+                @change=${this.onChange}
                 class="target-item"
+                supporting-text="required"
               ></md-outlined-text-field>
               <md-outlined-text-field
                 label="cluster"
+                name="Cluster"
+                type="number"
+                min="0"
+                max="65535"
+                @change=${this.onChange}
                 class="target-item"
+                supporting-text="optional"
               ></md-outlined-text-field>
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


Apollon77:
proposed text from @agners makes totally sense!

Apollon77 (changes requested):
So formally .. Please add the proposed text below the "add binding fields" in the popup. Thanks 

see https://github.com/matter-js/python-matter-server/pull/1242#issuecomment-3291534340

Luxni:
> @Luxni Can we not simply add your above note as an "note" below the text fields to tell that to the users?
> 
> Ideally the UI would read the descriptor cluster of the relevant endpoint and check which cluster clients are listed there and only allow to choose from these clusters (or "Multiple" when there is more than one entry), but that would be a bigger change, but we should add the note.

Understood. I'm trying to do this.

Apollon77:
@Luxni The message is enough for now, no need to do the cluster data analysis

Luxni:
> @Luxni The message is enough for now, no need to do the cluster data analysis

It is mainly to add a prompt information to prevent the wrong data input and cause problems with the device.

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
```diff
@@ -201,21 +201,40 @@ export class NodeBindingDialog extends LitElement {
...
       alert("Please enter a valid target endpoint");
       return;
     }
-    if (isNaN(targetCluster) || targetCluster < 0) {
-      alert("Please enter a valid target cluster");
-      return;
+
+    // cluster optional
+    if (targetCluster !== undefined) {
+      if (targetCluster < 0 || targetCluster > 65535) {
```
Apollon77, line 234:
```suggestion
      // We ignore vendor specific clusters for now
      if (targetCluster < 0 || targetCluster > 0x7fff) {
```

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
```diff
@@ -201,21 +201,40 @@ export class NodeBindingDialog extends LitElement {
...
+    ) {
       alert("Please enter a valid target node ID");
       return;
     }
-    if (isNaN(targetEndpoint) || targetEndpoint < 0) {
+
+    if (
+      targetEndpoint === undefined ||
+      targetEndpoint <= 0 ||
+      targetEndpoint > 65535
```
Apollon77, line 226:
```suggestion
      targetEndpoint > 0xfffe
```

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
```diff
@@ -201,21 +201,40 @@ export class NodeBindingDialog extends LitElement {
...
+      ? parseInt(this._targetNodeId.value, 10)
+      : undefined;
+    const targetEndpoint = this._targetEndpoint.value
+      ? parseInt(this._targetEndpoint.value, 10)
+      : undefined;
+    const targetCluster = this._targetCluster.value
+      ? parseInt(this._targetCluster.value, 10)
+      : undefined;
+
+    if (
```
Apollon77, line 214:
```suggestion
    // Matter Server does not use random NodeIds, so this is ok for now, but needs to be adjusted later
    if (
```

Apollon77: NodeIds are officially 64bit numbers, so ideally the textfield later should be a text and then you convert to BigInt. For now that works. When switching to matter.js finally I would like to go away from increasing node ids

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
```diff
@@ -298,17 +335,42 @@ export class NodeBindingDialog extends LitElement {
...
                 class="target-item"
+                @change=${this.onChange}
+                supporting-text="required"
               ></md-outlined-text-field>
               <md-outlined-text-field
                 label="endpoint"
+                name="Endpoint"
+                type="number"
+                min="0"
+                max="65535"
```
Apollon77, line 351:
```suggestion
                max="65534"
```

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
```diff
@@ -298,17 +335,42 @@ export class NodeBindingDialog extends LitElement {
...
+                @change=${this.onChange}
                 class="target-item"
+                supporting-text="required"
               ></md-outlined-text-field>
               <md-outlined-text-field
                 label="cluster"
+                name="Cluster"
+                type="number"
+                min="0"
+                max="65535"
```
Apollon77, line 361:
```suggestion
                max="32767"
```

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
```diff
@@ -298,17 +335,42 @@ export class NodeBindingDialog extends LitElement {
...
+                @change=${this.onChange}
                 class="target-item"
+                supporting-text="optional"
               ></md-outlined-text-field>
             </div>
+            <div style="margin:8px;">
+              <p style="font-size: 10px;font-style: italic;font-weight: bold;">
+                Note: When the Cluster is empty, the binding will be applied to
```
Apollon77, line 369: Please use the text proposed by @agners  Thanks

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
Luxni: What does `32767` mean of?

dashboard/src/components/dialogs/binding/node-binding-dialog.ts
Apollon77: I would still favor this because of the content. We will simplify that later again when matter.js is in

PR #1281, opened 2025-11-17
Update node.ts to handle non-string label data

mullender:
As reported in https://github.com/home-assistant/core/issues/156671 there is a case in which label is not a string but an object:
```
{
    "TLVValue": null,
    "Reason": "InteractionModelError: Failure (0x1)"
}
```
This then causes an error in the rendering of the dashboard.

Ideally we would figure out where the bad data comes from. But this should fix the symptom.


mullender:
I updated the pull request with the suggested change

mullender:
It looks like this needs a label. Perhaps `bugfix`

PR #1375, opened 2026-05-05
Handle DCL vendor info timeouts during startup

ivaschru:
## Summary

- Add an explicit 30 second timeout to the CSA DCL vendor-info fetch.
- Treat `TimeoutError` the same way as `ClientError` so vendor catalog refresh remains best-effort.
- Preserve startup by saving the built-in/stored vendor info instead of letting a stalled DCL response prevent the WebSocket server from binding.

## Why

In `MatterServer.start()`, `vendor_info.start()` runs before `mount_websocket()` and `MultiHostTCPSite.start()`. If the DCL vendor endpoint stalls until aiohttp raises `TimeoutError`, that exception is currently not handled and the Matter Server never reaches the WebSocket bind step. Supervisor can still report the add-on service as started, but `core-matter-server:5580/ws` keeps returning connection refused.

This matches the startup failure being tracked in `home-assistant/addons#4560`.

## Tests

```text
PYTHONPATH=. /tmp/ha_matter_debug/venv/bin/python -m pytest -o addopts='' tests/server/test_vendor_info.py tests/server/test_server.py -q
2 passed in 5.39s

/tmp/ha_matter_debug/venv/bin/python -m ruff check matter_server/server/vendor_info.py tests/server/test_vendor_info.py
All checks passed!
```

### DIFF
--- a/matter_server/server/vendor_info.py
+++ b/matter_server/server/vendor_info.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING
 
-from aiohttp import ClientError, ClientSession
+from aiohttp import ClientError, ClientSession, ClientTimeout
 
 from ..common.helpers.api import api_command
 from ..common.helpers.util import dataclass_from_dict, dataclass_to_dict
@@ -17,6 +17,7 @@ if TYPE_CHECKING:
 LOGGER = logging.getLogger(__name__)
 PRODUCTION_URL = "https://on.dcl.csa-iot.org"
 DATA_KEY_VENDOR_INFO = "vendor_info"
+DCL_REQUEST_TIMEOUT = ClientTimeout(total=30)
 
 
 TEST_VENDOR = VendorInfoModel(
@@ -71,7 +72,9 @@ class VendorInfo:
         LOGGER.info("Fetching the latest vendor info from DCL.")
         vendors: dict[int, VendorInfoModel] = {}
         try:
-            async with ClientSession(raise_for_status=True) as session:
+            async with ClientSession(
+                raise_for_status=True, timeout=DCL_REQUEST_TIMEOUT
+            ) as session:
                 page_token: str | None = ""
                 while page_token is not None:
                     async with session.get(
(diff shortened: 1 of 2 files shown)


ivaschru:
I validated this patch on the affected HAOS 17.2 / amd64 installation by building a temporary local add-on from the official `homeassistant/amd64-addon-matter-server:8.4.0` image with only `matter_server/server/vendor_info.py` replaced by this PR version.

Result from the live add-on startup:

```text
2026-05-05 19:50:27.898 INFO [matter_server.server.vendor_info] Fetching the latest vendor info from DCL.
2026-05-05 19:50:58.790 WARNING [matter_server.server.vendor_info] Unable to fetch vendor info from DCL:
TimeoutError
2026-05-05 19:50:58.793 INFO [matter_server.server.vendor_info] Saving vendor info to storage.
2026-05-05 19:50:58.793 INFO [matter_server.server.device_controller] Loaded 0 nodes from stored configuration
2026-05-05 19:50:58.824 INFO [matter_server.server.server] Matter Server successfully initialized.
```

And the listener is open afterwards:

```text
$ curl -sS -i --max-time 5 http://local-codex-matter-patch:5580/ws
HTTP/1.1 400 Bad Request
No WebSocket UPGRADE hdr: None
Can "Upgrade" only to "WebSocket".
```

So the patch fixes the startup blocker at the vendor-info DCL step: the server now reaches the WebSocket bind even when the DCL vendor endpoint times out.

One adjacent observation: on a fresh storage path, the earlier PAA DCL fetch can also hang until aiohttp's default ~5 minute timeout before falling back to Git. That is separate from this PR, but likely worth a follow-up timeout improvement for startup latency.

Apollon77:
Thanks a lot for your contribution. As the Readme states this project is no longer actively maintained in favor of the new Matter.js based Matter Server which is currently in Beta testing. There will be no new Python Matter Server release.

The Matter.js based server already have proper timeouts and is not blocking his start because of such timeouts, so this issue is already handled there.

That's why I close the PR, because it will be never released here anyway.

